PR #3, opened 2020-06-30
a basic recipe can be added to DataStore

isamitre:


src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -27,9 +17,39 @@
...
+    String recipeName = request.getParameter("recipeName");
+    String tagsResponse = request.getParameter("tags");
+    String ingredientsResponse = request.getParameter("ingredients");
+    String stepsResponse = request.getParameter("steps");
+    String privacy = request.getParameter("privacy");
 
-    response.sendRedirect("/pages/UserPage.html");
-  }
+    List<String> tags = new ArrayList<String>(Arrays.asList(tagsResponse.split(",", 0)));
+    List<String> ingredients = new ArrayList<String>(Arrays.asList(ingredientsResponse.split("\n", 0)));
```
ez-may: Some comments here and below to explain what you're doing and why would help in the future.

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -27,9 +17,39 @@
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // implement new recipe
+    String recipeName = request.getParameter("recipeName");
+    String tagsResponse = request.getParameter("tags");
+    String ingredientsResponse = request.getParameter("ingredients");
+    String stepsResponse = request.getParameter("steps");
+    String privacy = request.getParameter("privacy");
```
ez-may, line 25: It doesn't look like you use this anywhere, but I'm assuming that's later in the implementation?

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -27,9 +17,39 @@
...
+    }
+    for (String ingredient : ingredients) {
+      ingredient.trim();
+    }
+    for (String step : steps) {
+      step.trim();
+    }
 
-}
+    Entity recipeEntity = new Entity("PrivateRecipe");
```
ez-may, line 37: Will all recipes be saved as this type of entity later, or will you include an if-statement later to control the type of entity creation?

src/main/webapp/pages/UserPage.html
```diff
@@ -47,8 +47,8 @@ <h1 class="center">Create Recipe</h1>
             <!-- <option value="public">public</option> -->
             </select>
             <br><br>
-            <label for="image">Upload an image</label>
-            <input type="file"  accept="image/*" name="image" id="image">
+            <!-- <label for="image">Upload an image</label>
```
ez-may, line 51: Are you leaving this to keep working on it later?

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -27,9 +17,39 @@
...
+    String privacy = request.getParameter("privacy");
 
-    response.sendRedirect("/pages/UserPage.html");
-  }
+    List<String> tags = new ArrayList<String>(Arrays.asList(tagsResponse.split(",", 0)));
+    List<String> ingredients = new ArrayList<String>(Arrays.asList(ingredientsResponse.split("\n", 0)));
+    List<String> steps = new ArrayList<String>(Arrays.asList(stepsResponse.split("\n", 0)));
+
+    tags.removeAll(Arrays.asList("", null));
+    ingredients.removeAll(Arrays.asList("", null, "\n", "\r\n", "\r"));
```
ez-may, line 34: A comment here would make this understandable.

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: later when we have public recipes I'll have an if statement

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: yup once we have public recipes

src/main/webapp/pages/UserPage.html
isamitre: yeah I tried to make it work but it's really not lol so for the MVP I'll leave it and work on it again later

ez-may: Ok, then I would say add a comment with a to-do, and maybe the reason it's not MVP.

ez-may (approved):
LGTM

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -27,9 +17,41 @@
...
+
+    for (String tag : tags) {
+      tag.trim();
+    }
+    for (String ingredient : ingredients) {
+      ingredient.trim();
+    }
+    for (String step : steps) {
+      step.trim();
+    }
```
oturner24: I think you could do this as part of the splitting step above, and avoid iterating over all the lists. Here's a stackoverflow: https://stackoverflow.com/questions/41953388/java-split-and-trim-in-one-shot. You could do something like:
`List<String> tags = new ArrayList<String>(Arrays.asList(tagsResponse.trim().split("\\s*,\\s*");`
where you trim the original string, and then include whitespace in the split by using regular expressions. (if you're not familiar with regular expressions, here's a resource: https://medium.com/factory-mind/regex-tutorial-a-simple-cheatsheet-by-examples-649dc1c3f285. They're super handy for string parsing)

Or if you don't want to do that, take a look at Java 8 streams (see the second answer in that stackoverflow). It'll still be iterating over the lists but with much more concise (and in my opinion, more readable) syntax.

oturner24 (approved):

PR #5, opened 2020-07-01
Recipe rendering

ez-may:
Includes basic functionality to allow for the rendering of recipe information, and a new test file for viewing those changes. Updates were also made to fix issues with relative file path names.
image
image


src/main/webapp/javascript/RecipeRendering.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Takes the recipe info of a recipe as a JS object and the div that the recipe
+ * card will be added to. Creates only one recipe card. For the method to work,
+ * recipeInfo is expected to have the following data saved as properties,
+ * regardless of the explicit property name:
+ * a recipe name, recipe tags,  recipe description, and recipe ID to access more
+ * data on the recipe page.
+ */
+createRecipeCard = (divId, recipeInfo) => {
```
AlexanderEllis, line 9: I think this helper method is very helpful, but unfortunately I don't think your approach of building HTML this way is the right approach, given the potential for innerHTML misuse and XSS (https://en.wikipedia.org/wiki/Cross-site_scripting). For instance, if I created an object with the name `"<script>alert('hello')</script>"`, this would add that to the page as HTML and seems like it would call my script.

I'd encourage you to instead use those helper functions you mentioned, such as `document.createElement`. It may not be as short and sweet as your `createElement`, but it'll be much safer and you only have to write it once :)

src/main/webapp/javascript/RecipeRendering.js
```diff
@@ -0,0 +1,45 @@
...
+createElement = (object, htmlTag, tagOptions) => {
+    return "<" + htmlTag + " " + tagOptions + ">" + object + "</" + htmlTag + ">";
+}
+
+/**
+ * Method used for testing, currently used in test.jsp
+ * Use this to experiment with any loading functionality.
+ */
+load = () => {
+    createRecipeCard("content", {"name":"test", "description":"this is a test"});
```
AlexanderEllis: Do you want to extend this test a little bit? For example, it may be interesting to write some sort of test for this that guarantees a `p` element exists with the value `test` and class `recipe-card-name` and the same for the description.  That way you don't have to manually look at the results to check.

This could be something as simple as printing to the console (for now) or writing test results to the page (in test.jsp).

Additionally, if you write the tests before you refactor `createRecipeCard` to not use innerHTML, you could have the added confidence of knowing your tests still pass after you refactor :)

src/main/webapp/javascript/RecipeRendering.js
ez-may: So there is a lot more that will eventually go into this function to render the recipe card, so I wonder if it would it be possible to write a function to detect and encode html tags to display them as text? That way if a user writes HTML, it won't have any effect because it's being handled before it's displayed.
Edit: [Stack Overflow question](https://stackoverflow.com/questions/20855482/preventing-html-and-script-injections-in-javascript)

src/main/webapp/javascript/RecipeRendering.js
ez-may: Will do, I'll include a test in my next commit to be checked out.

ez-may: Test file is now included

src/main/webapp/javascript/RecipeRendering.js
ez-may: Attempt at editing the code as little as possible while still handling this problem has been included.

ez-may:
Note that in the test file, the object text is passed through the HTML remove function before being rendered, which seems redundant since the the HTML remove function is called in recipe rendering. The reason for this is to allow better debugging since the object text can be printed to the console, and also shows that the function is working as expected.
Images of testing functionality:
image
image

src/main/webapp/javascript/RecipeRendering.js
oturner24: I think it's awesome that you looked into and learned about XSS, but unfortunately, it's really difficult to guarantee that your own string manipulation will protect from any attack. Here's a stackoverflow that shows some examples that don't involve HTML tags: https://stackoverflow.com/questions/22395316/xss-without-html-tags (also here's a crazy long list: https://owasp.org/www-community/xss-filter-evasion-cheatsheet)

I know that in the limited way you're using the user entered text right now, it may be ok, but as you continue to add to this logic, it's possible that you'll open up vulnerabilities in your code, which will be harder to catch as the logic gets more complicated. You should also consider that other people may need to modify this code in the future, and may not have the same knowledge as you, and may unintentionally misuse it.

In general, with security stuff, you almost never want to implement a defense yourself. There's almost always a vetted solution that's a better guarantee than writing your own logic. So I do think you should use `document.createElement` and avoid using `innerHTML`. I know it's more writing, but it'll be safer, and like Alex said, you'll only have to write it once :)

That said, I do think the work/learning you've done is super valuable; knowing about XSS is super handy when doing web dev, so hopefully you don't feel like you've wasted any time. If you more questions, just let me know, I'm definitely happy to chat about this more.

ez-may: I've updated the code to now use JS DOM manipulation functions which don't implement inner HTML to show user inputted content, while still keeping an approach which allows for easy integration of new elements to the recipe cards.

ez-may:
Screenshot of the testing page with the most recent implementation.
image

src/main/webapp/javascript/tests.js
```diff
@@ -0,0 +1,145 @@
+/**
+ * TODO(matwsuaz): update with a new method to return refereces to live HTML DOM
+ * elements to reduce redundancy and code length.
+ * TODO(matwsuaz): update HTMl injection test to provide test result description
+ * and inform which test failed specifically.
+ */
+runTests = () => {
+    testCommentCard();
```
isamitre: should be testRecipeCard no?

src/main/webapp/javascript/tests.js
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * Makes a call to create a recipe card and makes sure the result is what is
+ * expected.
+ */
+testCommentCard = () => {
+    // The div in the test file for rendering the comment card test results
+    const testDiv = "recipe-card-test";
+
```
isamitre, line 18: nit: indentation should be +2 everywhere
[JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html#formatting-block-indentation)

src/main/webapp/pages/MainPage.jsp
```diff
@@ -5,7 +5,9 @@
     <title>Home</title>
   </head>
   <body>
-      <%@ include file="topbar.jsp"%>
-      <p> testing the default page</p>
+      <%@ include file="../components/topbar.jsp"%>
+      <div id="content">
+
```
isamitre, line 10: remove the extra empty line

isamitre (approved):

src/main/webapp/javascript/tests.js
ez-may: Ack and fixed

src/main/webapp/javascript/tests.js
ez-may: Ack and fixed

src/main/webapp/pages/MainPage.jsp
ez-may: ack and fixed

oturner24 (approved):
thanks for making these changes, I think the solution you ended with is great!

PR #7, opened 2020-07-01
Objects

isamitre:
made a User.java and udpated Recipe.java to include a description
updated UserPage.html and NewRecipeServlet.java to have the description implementation

### DIFF
--- a/src/main/java/com/google/sps/data/Recipe.java
+++ b/src/main/java/com/google/sps/data/Recipe.java
@@ -13,17 +13,28 @@ public class Recipe {
   private String recipeName;
   private long authorID;
   private String imageUrl;
+  private String description;
   private ArrayList<String> tags;
   private ArrayList<String> ingredients;
   private ArrayList<String> steps;
   private int popularity;
 
-  public Recipe(String recipeName, ArrayList<String> tags, ArrayList<String> ingredients, ArrayList<String> steps) {
+  public Recipe(long id, String recipeName, String description, ArrayList<String> tags, ArrayList<String> ingredients, ArrayList<String> steps) {
+    this.id = id;
     this.recipeName = recipeName;
     this.tags = tags;
     this.ingredients = ingredients;
     this.steps = steps;
     popularity = 0;
+  }
+
+  public Recipe(long id, String recipeName, String description, ArrayList<String> tags, ArrayList<String> ingredients, ArrayList<String> steps, int popularity) {
+    this.id = id;
+    this.recipeName = recipeName;
+    this.tags = tags;
+    this.ingredients = ingredients;
+    this.steps = steps;
+    this.popularity = popularity;
   }
 
   public void setImageUrl(String imageUrl) {
--- a/src/main/java/com/google/sps/data/Recipe.java
+++ b/src/main/java/com/google/sps/data/Recipe.java
@@ -7,7 +7,7 @@ import java.util.List;
 public class Recipe {
   private long id;
   private String recipeName;
-  private long authorID;
+  private String authorID;
   // image is not implemented yet
   private String imageUrl;
   private String description;
@@ -17,18 +17,7 @@ public class Recipe {
   private int popularity;
 
   public Recipe(long id, String recipeName, String description, ArrayList<String> tags, 
-  ArrayList<String> ingredients, ArrayList<String> steps) {
-    this.id = id;
-    this.recipeName = recipeName;
-    this.description = description;
-    this.tags = tags;
-    this.ingredients = ingredients;
-    this.steps = steps;
-    popularity = 0;
-  }
-
-  public Recipe(long id, String recipeName, String description, ArrayList<String> tags, 
-  ArrayList<String> ingredients, ArrayList<String> steps, int popularity) {
+      ArrayList<String> ingredients, ArrayList<String> steps, int popularity) {
     this.id = id;
     this.recipeName = recipeName;
     this.tags = tags;
@@ -45,7 +34,7 @@ public class Recipe {
     this.id = id;
   }
   
-  public void setAuthorID(long id) {
+  public void setAuthorID(String id) {
     authorID = id;
   }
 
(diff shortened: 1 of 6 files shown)


isamitre:
image
image

isamitre:
image

Aliblas (approved):
Looks good to me. I think it would be good to have a screenshot of the recipe post request here too, since there were changes made to that.

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,57 @@
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+// Class representing a User
+public class User {
+  private long id;
```
Aliblas: user ID will have to be changed to a String, google sign-in API's getID method returns as an opaque String.

Aliblas (approved):

isamitre:
image

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -2,30 +2,41 @@
...
+    this.id = id;
     this.recipeName = recipeName;
+    this.description = description;
     this.tags = tags;
     this.ingredients = ingredients;
     this.steps = steps;
     popularity = 0;
   }
 
+  public Recipe(long id, String recipeName, String description, ArrayList<String> tags,
```
oturner24: Do you need to set the description below?

Also, do you need 2 separate constructors here and above? They seem the same except for this one takes popularity as a parameter. Could you delete the one above, and just use this one, when you don't have popularity you can just pass 0?

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,57 @@
+package com.google.sps.data;
+
+import java.util.ArrayList;
+import java.util.List;
+
+// Class representing a User
+public class User {
+  private String id;
```
oturner24, line 8: Why is the user id a string but the recipe id and author id a long?

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,57 @@
...
+  public User(String id, String email, String username) {
+    this.id = id;
+    this.email = email;
+    this.username = username;
+    planner = new ArrayList<Long>();
+    cookbook = new ArrayList<Long>();
+    userRecipes = new ArrayList<Long>();
+  }
+
+  public User(String id, String email, String username, ArrayList<Long> planner, ArrayList<Long> cookbook, ArrayList<Long> userRecipes) {
```
oturner24: When are you going to be using this constructor? Won't a new user always start with empty lists?

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public void setID(String id) {
+    this.id = id;
+  }
+
+  public boolean equalIDs(String id) {
+    if (this.id.equals(id)) {
+      return true;
+    }
+    return false;
```
oturner24: you can just do: `return this.id.equals(id)`

Alternatively, since this operation is super simple, maybe it would make more sense to have a getID method and then wherever you're calling this from can check if the IDs are equal

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,57 @@
...
+    planner.add(id);
+  }
+
+  public void addRecipeToCookbook(long id) {
+    cookbook.add(id);
+  }
+
+  public void addRecipeToUserRecipes(long id) {
+    userRecipes.add(id);
+  }
```
oturner24, line 44: You probably also want `removeRecipeFromPlanner` and so on, right?

oturner24 (changes requested):

src/main/java/com/google/sps/data/User.java
isamitre: because Ali says that the authentication API gives the user a String id but DataStore gives you a long id
the author id should be a String tho I'll fix that

src/main/java/com/google/sps/data/Recipe.java
isamitre: yeah I'll do that

isamitre: what do you mean by set the description though?

src/main/java/com/google/sps/data/User.java
isamitre: it feels sketchy to return the user ID though
shouldn't that be private? or does it not really matter?

src/main/java/com/google/sps/data/User.java
isamitre: lol yeah can't believe I forgot that

src/main/java/com/google/sps/data/Recipe.java
oturner24: Oh I just meant I think you're missing `this.description = description`?

src/main/java/com/google/sps/data/User.java
oturner24: Oh that's an interesting point. I guess it depends on whether you think the user ID is sensitive, which in my opinion depends on where/how you're accessing it and what is the potential danger of it being exposed (although I'm not a security expert). I think it's fine to leave this as an equals method in that case.

src/main/java/com/google/sps/data/Recipe.java
isamitre: oh lol yeah

src/main/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public void removeRecipeFromPlanner(long id) {
+    planner.remove(new Long(id));
+  }
+
+  public void removeRecipeFromCookbook(long id) {
+    cookbook.remove(new Long (id));
```
oturner24: nit: remove the space between Long and ( (here and below)

oturner24 (approved):

PR #8, opened 2020-07-01
Delete recipe

isamitre:
beginning of the backend for deleting a recipe

### DIFF
--- a/src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
+++ b/src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
@@ -13,7 +13,7 @@ import javax.servlet.http.HttpServletResponse;
 
 /** Servlet responsible for deleting recipes. */
 @WebServlet("/delete-recipe")
-public class DeleteTaskServlet extends HttpServlet {
+public class DeleteRecipeServlet extends HttpServlet {
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
--- a/src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
+++ b/src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
@@ -1,0 +1,26 @@
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Entity;
+import java.io.IOException;
+import java.util.*;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting recipes. */
+@WebServlet("/delete-recipe")
+public class DeleteTaskServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long id = Long.parseLong(request.getParameter("id"));
+
+    Key taskEntityKey = KeyFactory.createKey("PrivateRecipe", id);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.delete(taskEntityKey);
+  }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -0,0 +1,26 @@
...
+import java.io.IOException;
+import java.util.*;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting recipes. */
+@WebServlet("/delete-recipe")
+public class DeleteTaskServlet extends HttpServlet {
```
Aliblas: Class name should be same as filename: DeleteRecipeServlet

Aliblas, line 16: Class name should be same as filename:  DeleteRecipeServlet

src/main/webapp/javascript/user-page.js
```diff
@@ -0,0 +1,16 @@
...
+function closeModal() {
+  document.getElementById("recipe-modal").style.display = "none";
+}
+
+// Tells the server to delete the recipe.
+// Input is a Recipe
+function deleteRecipe(recipe) {
+  const params = new URLSearchParams();
+  params.append('id', recipe.id);
+  fetch('/delete-recipe', {method: 'POST', body: params});
```
Aliblas: I think double quotes are used here, correct me if I'm wrong.

Aliblas (changes requested):

src/main/webapp/javascript/user-page.js
isamitre: idk I copied it from the todo list example

isamitre: I'll change it to double quotes tho

Aliblas (approved):

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -0,0 +1,26 @@
...
+
+/** Servlet responsible for deleting recipes. */
+@WebServlet("/delete-recipe")
+public class DeleteRecipeServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    long id = Long.parseLong(request.getParameter("id"));
+
+    Key taskEntityKey = KeyFactory.createKey("PrivateRecipe", id);
```
oturner24, line 23: Can only private recipes be deleted? If so, maybe add a comment here with a little explanation on why you made that design decision

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -0,0 +1,26 @@
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    datastore.delete(taskEntityKey);
```
oturner24, line 25: Does datastore handle delete gracefully? ie what happens if it can't find the key to delete?

oturner24 (approved):

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
isamitre: right now private recipes are user uploaded recipes; public recipes right now are already part of the website, no one can delete them
eventually we'll let users publish their recipes and make them public but that's for after the MVP

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
isamitre: I don't know and the documentation doesn't explicitly say
stackoverflow says that you can make a key for an entity even if the entity doesn't exist, so if I do so then delete the key I think it should be fine?

oturner24: Looks like it might throw an IllegalArgumentException: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/DatastoreService.html#delete

You'll probably want to check that the key exists first before trying to delete it (I know you already merged this so feel free to do this in your next PR)

PR #9, opened 2020-07-06
Auth/refactor login logout

Aliblas:
Fully functioning authentication system. Servlets now return Json instead of redirecting. UserAuth.js has been updated to accommodate these changes.

### DIFF
--- a/src/main/java/com/google/sps/servlets/UserStatusServlet.java
+++ b/src/main/java/com/google/sps/servlets/UserStatusServlet.java
@@ -35,15 +35,12 @@ public class UserStatusServlet extends HttpServlet {
   @Override
   public void doGet (HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("application/json");
-    UserService userService = UserServiceFactory.getUserService();
 
     JsonObject userInfo = new JsonObject();
-    userInfo.addProperty("isUserLoggedIn", userService.isUserLoggedIn());
-    if (userInfo.get("isUserLoggedIn").getAsBoolean()) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/servlets/UserStatusServlet.java
```diff
@@ -33,17 +33,14 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     response.setContentType("application/json");
-    UserService userService = UserServiceFactory.getUserService();
 
     JsonObject userInfo = new JsonObject();
-    userInfo.addProperty("isUserLoggedIn", userService.isUserLoggedIn());
-    if (userInfo.get("isUserLoggedIn").getAsBoolean()) {
+    userInfo.addProperty("isUserLoggedIn", UserHelper.isUserLoggedIn());
+    if (UserHelper.isUserLoggedIn()) {
         System.out.println("User is logged in");
```
isamitre, line 42: is this if statement supposed to be here or was it just a debugging thing?

isamitre (approved):

src/main/java/com/google/sps/servlets/UserStatusServlet.java
Aliblas: I might return info from the google sign-in API's user profile some time in the future, so this if statement is a place-holder.

Aliblas: I've changed it to reflect that

oturner24 (approved):

PR #10, opened 2020-07-06
Upload image

isamitre:
someone can upload an image on the recipe form and it's blobkey will be saved in DataStore as a String with the rest of the recipe


src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -8,8 +8,7 @@
   private long id;
```
ez-may: This is a general Nit, but could you try to keep class attributes alphabetical?

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -27,19 +26,15 @@ public Recipe(long id, String recipeName, String description, ArrayList<String>
     this.popularity = popularity;
   }
 
-  public void setImageUrl(String imageUrl) {
-    this.imageUrl = imageUrl;
+  public void setImageBlobKey(String blobKey) {
```
ez-may, line 29: I there a reason why this is a separate method to add the image blob key, but not part of the constructor?

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -27,19 +26,15 @@ public Recipe(long id, String recipeName, String description, ArrayList<String>
...
-    this.id = id;
-  }
-  
-  public void setAuthorID(String id) {
+  public void setAuthorId(String id) {
     authorID = id;
   }
 
-  public long getID() {
+  public long getId() {
```
ez-may, line 37: Will other get methods be included in the future?

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -35,12 +43,29 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     ingredients.removeAll(Arrays.asList("", null, "\n", "\r\n", "\r"));
     steps.removeAll(Arrays.asList("", null, "\n", "\r\n", "\r"));
 
+    BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
```
ez-may, line 48: a comment here to explain what you are doing

src/main/java/com/google/sps/servlets/UploadServlet.java
```diff
@@ -0,0 +1,21 @@
...
+
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import javax.servlet.annotation.WebServlet;
+
```
ez-may, line 11: A comment here to explain what this servlet for

src/main/webapp/javascript/user-page.js
```diff
@@ -1,6 +1,18 @@
 
 function openModal() {
   document.getElementById("recipe-modal").style.display = "block";
+  fetchBlobstoreUrl();
+}
+
```
ez-may, line 6: Comment documentation for your method

ez-may (approved):

src/main/java/com/google/sps/data/Recipe.java
isamitre: because I can't actually render an image right now so I didn't want to make it mandatory for making a Recipe
do you think I should just make it part of the constructor for when I eventually get it working?

src/main/java/com/google/sps/data/Recipe.java
isamitre: yeah anything we might need this is by no means done
why, what do you want me to add?

src/main/java/com/google/sps/data/Recipe.java
isamitre: yeah sure

src/main/webapp/javascript/user-page.js
isamitre: it's not that I'm ignoring this comment, but I'm not exactly sure what this method does
it was in the blobstore API tutorial
I'll try to figure it out for my next PR

src/main/java/com/google/sps/data/Recipe.java
ez-may: That makes sense. I think a TODO or comment for yourself so you don't accidentally forget when you do get it running would be helpful.

src/main/java/com/google/sps/data/Recipe.java
ez-may: Nothing in particular, I was more thrown off there was only one get method in this PR so I wanted ask about it, and whether you wanted to go ahead and add them all or just leave it for later.

oturner24 (approved):

PR #11, opened 2020-07-08
List user recipes

isamitre:
backend for listing private recipes is done
backend for images is done
there's a createImage method for the frontend in RecipeRendering

### DIFF
--- a/src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
@@ -24,6 +24,7 @@ public class ListPrivateRecipesServlet extends HttpServlet {
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     // this Query seems to need the addSort method to work so rn it's arbitrarily by the recipeName
     // it can be changed
+    // TODO: filter by author ID
     Query query = new Query("PrivateRecipe").addSort("recipeName", SortDirection.DESCENDING);
  
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
--- a/src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
@@ -1,0 +1,55 @@
+package com.google.sps.servlets; 
+ 
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.gson.Gson;
+import com.google.sps.data.Recipe;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+ 
+/** Servlet responsible for listing private recipes. */
+@WebServlet("/list-private-recipes")
+public class ListPrivateRecipesServlet extends HttpServlet {
+   
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    Query query = new Query("PrivateRecipe").addSort("recipeName", SortDirection.DESCENDING);
+ 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+ 
+    List<Recipe> recipes = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+    //   TODO: check if logged in user's id == recipe's authorID
+    //   String authorID = (String) entity.getProperty("authorID");
+      long id = entity.getKey().getId();
+      String recipeName = (String) entity.getProperty("recipeName");
+      String description = (String) entity.getProperty("description");
+      String blobkey = (String) entity.getProperty("imageBlobKey");
+      ArrayList<String> tags = (ArrayList<String>) entity.getProperty("tags");
+      ArrayList<String> ingredients = (ArrayList<String>) entity.getProperty("ingredients");
+      ArrayList<String> steps = (ArrayList<String>) entity.getProperty("steps");
+
+      Recipe recipe = new Recipe(id, recipeName, description, tags, ingredients, steps, 0);
+      recipe.setImageBlobKey(blobkey);
+      recipes.add(recipe);
+    }
+    System.out.println("        RECIPE SIZE    " + recipes.size());
+ 
+    Gson gson = new Gson();
+
+    response.setContentType("application/json;");
+    response.getWriter().println(gson.toJson(recipes));
+  }
+}
+ 
+
--- a/src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
+++ b/src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
@@ -22,10 +22,10 @@ public class ListPrivateRecipesServlet extends HttpServlet {
    
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    // this Query seems to need the addSort method to work so rn it's arbitrarily by the recipeName
+    // this Query seems to need the addSort method to work so rn it's arbitrarily by the name
     // it can be changed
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
```diff
@@ -0,0 +1,55 @@
...
+ 
+/** Servlet responsible for listing private recipes. */
+@WebServlet("/list-private-recipes")
+public class ListPrivateRecipesServlet extends HttpServlet {
+   
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // this Query seems to need the addSort method to work so rn it's arbitrarily by the recipeName
+    // it can be changed
+    Query query = new Query("PrivateRecipe").addSort("recipeName", SortDirection.DESCENDING);
```
ez-may: Are you planning on implementing a search for all private recipes and then filtering it? After reading the datastore API, I'm pretty sure there are ways to filter the entities by author ID before getting the data and having to filter it on the Front End.

src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
```diff
@@ -0,0 +1,55 @@
...
+ 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(query);
+ 
+    List<Recipe> recipes = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
```
ez-may, line 34: This looks fine, but if you're only going to render recipe cards with this info, you may only need one image, the recipe name, author, datastore ID, and the description. Steps, tags, ingredients, etc. can be left until the recipe is shown on the actual webpage.

src/main/webapp/javascript/user-page.js
```diff
@@ -1,9 +1,29 @@
+// loads all the recipes when the recipe loads
+async function loadRecipes() {
```
ez-may, line 2: Is there a reason there are two functions instead of just one which loads all the recipes?

src/main/webapp/javascript/user-page.js
```diff
@@ -1,9 +1,29 @@
...
+  loadUserRecipes();
+}
+
+async function loadUserRecipes() {
```
ez-may, line 6: Is this currently loading user recipes correctly? It looks like you've commented out the part for rendering.

src/main/webapp/javascript/RecipeRendering.js
```diff
@@ -45,4 +45,14 @@ createElement = (htmlTag, object = "", tagOptions = {}) => {
...
     return htmlElement;
-}
\ No newline at end of file
+}
+
+/**
+ * Takes two objects which represent the recipeName and the blobkey
+ * Returns the reference to the constructed image element. 
+ */
+createImage = (recipeName, blobkey) => {
```
ez-may: Why does this take a recipeName and then not use it?

src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
isamitre: oh that's a good point

src/main/webapp/javascript/user-page.js
isamitre: yeah it's not rendering yet

src/main/webapp/javascript/RecipeRendering.js
isamitre: lol I forgot! I was going to make the recipe name the alt text

ez-may: Sounds good!

src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
isamitre: sounds good! I'll do that once I have the authorID implemented

src/main/webapp/javascript/user-page.js
isamitre: no real reason I just think it'll be neater since I'll have to call different servlets and stuff

src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
isamitre: I'll change it in the next PR tho

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -25,13 +25,21 @@
...
     String stepsResponse = request.getParameter("steps");
     // privacy will be used once we give users the option to make their recipes public
-    String privacy = request.getParameter("privacy");
+    // String privacy = request.getParameter("privacy");
+
+    // prevents empty responses as being saved in datastore
+    // redirects user back to UserPage
+    // TODO: inform the user that they're missing stuff
+    if (name.equals("") || tagsResponse.equals("") || description.equals("") || ingredientsResponse.equals("") || stepsResponse.equals("")){
+      response.sendRedirect("/pages/UserPage.html");
```
ez-may, line 40: You added some redirecting here which wasn't in the last review. For now I say it looks good, but Ali and I have both had instances where this doesn't work as expected, so I would suggest double checking it behaves properly.

ez-may (approved):

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: I've played around with it and it seems to work but if it starts breaking out of nowhere I'll know to look here

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -25,13 +25,21 @@
...
+    String name = request.getParameter("name");
     String tagsResponse = request.getParameter("tags");
     String description = request.getParameter("description");
     String ingredientsResponse = request.getParameter("ingredients");
     String stepsResponse = request.getParameter("steps");
     // privacy will be used once we give users the option to make their recipes public
-    String privacy = request.getParameter("privacy");
+    // String privacy = request.getParameter("privacy");
+
+    // prevents empty responses as being saved in datastore
```
oturner24, line 36: I think you could also do this by making the form fields required: https://www.w3schools.com/tags/att_input_required.asp

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -25,13 +25,21 @@
...
+    // redirects user back to UserPage
+    // TODO: inform the user that they're missing stuff
+    if (name.equals("") || tagsResponse.equals("") || description.equals("") || ingredientsResponse.equals("") || stepsResponse.equals("")){
```
oturner24, line 39: Do you also want to trim these strings?

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -43,37 +51,37 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    recipeEntity.setProperty("steps", steps);
+
     // getUploads returns a set of blobs that have been uploaded 
     // the Map object is a list that associates the names of the upload fields to the blobs they contained
     BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
     Map<String, List<BlobKey>> blobs = blobstoreService.getUploads(request);
     List<BlobKey> blobKeys = blobs.get("image");
-    // TODO: make sure blobkey exists before initializing it
-    BlobKey blobkey = blobKeys.get(0);
     boolean noImage = false;
```
oturner24: look like you can remove this boolean now?

oturner24 (approved):

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: I tried doing that (all the form fields are required) but it still saved in DataStore with empty inputs

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: yeah probably I'll do that

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: yup!

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
oturner24: It looks like you put `required` on the <label>, but I think it needs to be on the <input>. If it still doesn't work, it's fine to do it this way too, it just seems like it'd be easier to handle on the frontend.

PR #13, opened 2020-07-09
Render user recipe cards

isamitre:
when there are no recipes: 
image

images automatically resize to be squares
recipe cards adjust with the size of the window, appearing in three, two, and one column(s) as needed

I'll put the screenshot of it with recipes later I have technical difficulties rn lol


Aliblas (approved):
This looks all good, I can find nothing to nitpick. I hope we can discuss merging this with the sign-in API features come morning.

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -15,6 +15,14 @@
   private int popularity;
   private long id;
 
+  // Recipe constructor for the recipe cards for which less information is needed
+  public Recipe(long id, String name, String imageBlobKey, String description) {
```
oturner24, line 19: You also may want to initialize the array lists in this constructor (just to empty lists). Even though you won't be using them, it's a good idea to have all your class variables initialized so you don't run into null pointer exceptions later on

src/main/java/com/google/sps/servlets/ListPrivateRecipesServlet.java
```diff
@@ -24,25 +24,21 @@
...
       long id = entity.getKey().getId();
       String name = (String) entity.getProperty("name");
       String description = (String) entity.getProperty("description");
       String blobkey = (String) entity.getProperty("imageBlobKey");
-      ArrayList<String> tags = (ArrayList<String>) entity.getProperty("tags");
-      ArrayList<String> ingredients = (ArrayList<String>) entity.getProperty("ingredients");
-      ArrayList<String> steps = (ArrayList<String>) entity.getProperty("steps");
 
-      Recipe recipe = new Recipe(id, name, blobkey, description, tags, ingredients, steps, 0);
+      Recipe recipe = new Recipe(id,  name,  blobkey,  description);
```
oturner24: nit: looks like there are 2 spaces between parameters.

src/main/webapp/css/style-userpage.css
```diff
@@ -3,6 +3,39 @@
   display: flex;
   height: 30px;
   justify-content: space-between;
+  margin: 5px 0;
+  /* min-width: 530px; */
```
oturner24: remove this?

src/main/webapp/css/style.css
```diff
@@ -13,3 +17,81 @@
...
+
+.image-div {
+  position: relative;
+}
+.recipe-card-image {
+  width: 24vw;
+  height: 24vw;
+  overflow: hidden;
+  object-fit: cover;
+  /* border-radius: 10px; */
```
oturner24, line 45: remove this? Or is this related to the TODO above?

src/main/webapp/css/style.css
```diff
@@ -13,3 +17,81 @@
...
+}
+@media screen and (min-width: 1250px) {
+  .recipe-card {
+    width: 300px;
```
oturner24: nit: alphabetize, here and a couple other places width is out of order

src/main/webapp/javascript/user-page.js
```diff
@@ -7,16 +8,65 @@ async function loadUserRecipes() {
...
   if (Object.keys(recipes)) {
-    const recipesElement = document.getElementById('user-recipes');
-    recipesElement.innerText = "";
     for (let key of Object.keys(recipes)) {
+        size++;
         let value = recipes[key];
-        // createUserRecipeCard(value)
+        createUserRecipeCard(value);
+    }
+    if(size == 0){
```
oturner24: nit: space after if and before {

also, you may be able to check Object.keys(recipes).length to get the size, instead of keeping track of a size variable.

src/main/webapp/javascript/user-page.js
```diff
@@ -7,16 +8,65 @@ async function loadUserRecipes() {
...
+  elementsToAddToImageDiv.forEach(elem => imageDiv.appendChild(elem));
+  elementsToAddToTextDiv.forEach(elem => textDiv.appendChild(elem));
+  allElementsToAdd.forEach(elem => recipeDiv.appendChild(elem));
+  docDiv.appendChild(recipeDiv);
+
+  let plannerButtons = document.getElementsByClassName("planner-btn");
+  let cookbookButtons = document.getElementsByClassName("cookbook-btn");
+  const add1 = createElement("i", "add_circle_outline", {"class": "material-icons"});
+  const add2 = createElement("i", "add_circle_outline", {"class": "material-icons"});
+  for (let i = 0; i < plannerButtons.length; i++) {
```
oturner24: Will plannerButtons and cookbookButtons always be the same size? If so, can you add a comment like "using plannerButtons.length is ok because plannerButtons and cookbookButtons will always be the same length"? Even better, consider adding a check to verify that before this loop.

src/main/webapp/pages/UserPage.html
```diff
@@ -2,24 +2,31 @@
...
+        </div>
+        <div class="recipes-container curve" id="planner-recipes">
         </div>
       </div>
       <!-- Cookbook content -->
       <div>
-        <h2>Cookbook</h3>
-        <div class="recipes-container">
+        <div id="header">
+          <h2>Cookbook</h3>
```
oturner24: should open tag and close tag both be h2?

oturner24 (changes requested):

isamitre:
image

src/main/webapp/css/style.css
isamitre: related to the TODO

src/main/webapp/javascript/user-page.js
isamitre: oh okay that makes sense 
I tried to get the size of recipes by itself and it didn't work which is why I did the size thing

src/main/webapp/javascript/user-page.js
isamitre: yeah there will be as many plannerButtons/cookbookButtons as there are recipe cards
will do!

src/main/webapp/pages/UserPage.html
isamitre: lol yeah

src/main/webapp/css/style-userpage.css
```diff
@@ -3,6 +3,43 @@
...
+    width: 73vw;
+    margin: 3vw;
+    min-width: 280px;
+  }
+  .recipe-card-image {
+    width: 71vw;
+    height: 71vw;
+    min-width: 280px;
+    min-height: 280px;
+  }
```
oturner24: nit: alphabetize the properties

src/main/webapp/javascript/user-page.js
```diff
@@ -7,16 +8,64 @@ async function loadUserRecipes() {
   const response = await fetch('/list-private-recipes');
   const recipes = await response.json();
 
+  let recipesDiv = document.getElementById("user-recipes");
+  recipesDiv.innerHTML = "";
+  
+  let size = 0;
```
oturner24: think you can remove this now

src/main/webapp/javascript/user-page.js
```diff
@@ -7,16 +8,64 @@ async function loadUserRecipes() {
...
     for (let key of Object.keys(recipes)) {
+        size++;
         let value = recipes[key];
-        // createUserRecipeCard(value)
+        createUserRecipeCard(value);
+    }
+    if(Object.keys(recipes).length == 0){
+        recipesDiv.innerText = "You have not uploaded any recipes yet.";
+        recipesDiv.style.height = "100px";
     }
```
oturner24: sort of a nit, but I would do this before iterating through the keys, the result is the same but I feel like it makes more sense/makes it more readable to check the length first, ie:

if (length == 0) 
  // update recipesDiv
else 
  // iterate through keys and create recipe cards

since these are actually mutually exclusive, ie only one of the 2 should ever happen, never both

src/main/webapp/javascript/user-page.js
```diff
@@ -7,16 +8,64 @@ async function loadUserRecipes() {
...
   }
 }
 
+createUserRecipeCard = (recipeInfo) => {
```
oturner24: this seems really similar to createRecipeCard in RecipeRendering.js - do you need both?

src/main/webapp/javascript/RecipeRendering.js
```diff
@@ -6,7 +7,7 @@
  * a recipe name, recipe tags,  recipe description, and recipe ID to access more
  * data on the recipe page.
  */
-createRecipeCard = (divId, recipeInfo) => {
+createRecipeCardTest = (divId, recipeInfo) => {
```
oturner24: In general, you should put tests in their own file. This doesn't seem like a true test though, more like a first draft? Can you remove it now that you have it working?

isamitre:
heads up: I changed how I render my recipe cards a bit by removing the specific createUserRecipeCard and instead using createRecipeCard and adding the edit/delete buttons after the fact

oturner24 (approved):

src/main/webapp/javascript/RecipeRendering.js
isamitre: yeah it was Mathew's original function so I felt bad deleting it but I suppose I should

src/main/webapp/javascript/user-page.js
isamitre: no I don't and I've changed it

PR #14, opened 2020-07-10
style fixes, comments improved, check for user before inserting user …

Aliblas:
This is the sign in interface when sign out:
Screenshot 2020-07-10 at 10 50 27 AM
This is the interface when signed in:
Screenshot 2020-07-10 at 10 49 53 AM
And this is the interface when the user hovers over it:
Screenshot 2020-07-10 at 10 50 03 AM
Currently default pfp is the chicken.

### DIFF
--- a/src/main/webapp/javascript/user-auth.js
+++ b/src/main/webapp/javascript/user-auth.js
@@ -44,7 +44,7 @@ function initSigninV2() {
 }
 
 /**
- * Signs our current user from g-signin API.
+ * Signs out current user from g-signin API.
  */
 function signOut() {
   auth2.signOut();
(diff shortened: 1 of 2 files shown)


src/main/webapp/javascript/user-auth.js
```diff
@@ -56,7 +62,7 @@ function signOut() {
  * @param {GoogleUser} current user to provide profile info.
  */
 function insertUserInfo(googleUser) {
-  // insert user name into user-page anchor
+  // Insert user name into user-page anchor
```
AlexanderEllis: Super nit: period at end of sentence

(mostly so we can try adding a new commit to see if it undoes approval)

AlexanderEllis (approved):
Thanks for the fixes! LGTM

isamitre (approved):

PR #15, opened 2020-07-10
Delete recipe

isamitre:
recipes can be deleted and it prompts the user to confirm they wish to delete the recipe before deleting it
image


src/main/webapp/javascript/user-page.js
```diff
@@ -30,10 +29,38 @@ async function loadUserRecipes() {
           elementsToAddToImageDiv.forEach(elem => imageDivs[i].appendChild(elem));
         }
       }
+
+      // functionality of delete button
```
ez-may: I think this delete logic could merit its own method, and then only call it in loadUserRecipes(). This could help in the future when you add functionality to handle adding to cookbook or planner, where having all the functionality in one method could become difficult to understand.

src/main/webapp/javascript/user-page.js
```diff
@@ -30,10 +29,38 @@ async function loadUserRecipes() {
...
+      const deleteButtons = document.getElementsByClassName('fa-trash-alt');
+      const recipeCards = document.getElementsByClassName('recipe-card');
+      // there are as many delete buttons as there are recipe cards
+      for (let i = 0; i < deleteButtons.length; i++) {
+        let recipe = recipes[i];
+        let recipeCard = recipeCards[i];
+        deleteButtons[i].addEventListener('click', () => {
+          const deleteConfirmed = confirm("Are you sure you want to delete the " 
+            +recipe.name+" recipe?\nThis action cannot be undone!");
```
ez-may: Nit: space between operators ("+") and other parts of the expression.

src/main/webapp/javascript/user-page.js
```diff
@@ -12,8 +12,7 @@ async function loadUserRecipes() {
   
   if (Object.keys(recipes)) {
     if(Object.keys(recipes).length == 0){
```
ez-may: Nit: spaces for the if statement and bracket are missing

src/main/webapp/javascript/user-page.js
```diff
@@ -30,10 +29,42 @@ async function loadUserRecipes() {
           elementsToAddToImageDiv.forEach(elem => imageDivs[i].appendChild(elem));
         }
       }
+      addDeleteFunctionality(recipes);
+      // TODO: add edit functionality
     }
   }
 }
 
+function addDeleteFunctionality(recipes){
```
ez-may, line 45: Need to add comment for function

src/main/webapp/javascript/user-page.js
```diff
@@ -30,10 +29,42 @@ async function loadUserRecipes() {
...
+      }
+      if (recipeCards.length == 0) {
+        noRecipes("user-recipes", "You have not uploaded any recipes yet.");
+      }
+    });
+  }
+}
+
+// takes in div id and message
+noRecipes = (divID, message) => {
```
ez-may: I know what this does, but I think that having an expanded comment for this function could improve readability in the future regarding use.

ez-may (approved):

src/main/webapp/javascript/user-page.js
isamitre: isn't it pretty obvious based on the function name and parameter?

src/main/webapp/javascript/user-page.js
```diff
@@ -30,10 +35,44 @@ async function loadUserRecipes() {
...
+        noRecipes("user-recipes", NO_USER_RECIPES);
+      }
+    });
+  }
+}
+
+// takes in div id and message
+// makes the recipes container bigger and gives it the message
+// to use when there are no recipes in planner/cookbook/user-recipes
+noRecipes = (divID, message) => {
```
oturner24: Can you give this a more descriptive name? `noRecipes` indicates the situation but not what the function is actually doing, something like `setupDivWithNoRecipes` or `setNoRecipesMessage` I think would be more clear

oturner24 (approved):

src/main/webapp/javascript/user-page.js
isamitre: okay sounds good!

PR #16, opened 2020-07-13
Search

ez-may:
This PR is the basic implementation for the search loop, and still doesn't include features for redirecting, styling results, regular expressions for search tags, and more. This PR introduces the following:
--> A new component called GlobalConfigs to begin including JS and CSS files that should be present in every part of the website to standardize behavior.
--> A new search bar implemented with JS since HTML forms don't allow for the desired functionality
--> Search servlet for handling requests and helper classes with functions to implement basic search loop. 
--> Results from the servlet are rendered poorly on the page that made the request. There is no current redirecting functionality.
--> Results are not cleared when a user makes a new query
Current UI result:
Screenshot 2020-07-15 at 12 05 54 PM


src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public UserQuery(String query) {
+    String standardizedQuery= SearchUtils.standardize(query);
+    this.tagsAndIngredients = new ArrayList<String>();
+    this.authors = new ArrayList<String>();
+    this.recipeName = SearchUtils.parseRecipeName(standardizedQuery);
+  }
+
+  /**
+   * Takes a new tag or ingeredient to search for and adds it to the tagsAndIngredients attribute.
```
Aliblas: nit: Typo

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+   * TODO: Implement ability to avoid duplicates
+   */
+  public void addTag(String tag) {
+    this.tagsAndIngredients.add(tag);
+  }
+
+  public ArrayList<String> getTagsAndIngreds() {
```
Aliblas: Nit: Longer function name is preferable to truncated, at least in this case.

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** 
+ * Handles requests for a Datastore search based on a query string with filters to use.
+ **/
+@WebServlet("/search")
+public class Search extends HttpServlet {
```
Aliblas: As per convention, consider naming this file SearchServlet.java, and the servlet SearchServlet.

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Hnadles the get request for Datastore search results. Processes the url search parameters
```
Aliblas: Nit: Typo.

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Initializes a user query object which will seperate the query data for ease of access.
+    UserQuery uQuery = new UserQuery(request.getParameter("query"));
+    
+    // Creates the new Query and adds the filter to be used in Datastore
+    // The filter is currently only set to find an exact match name in Datastore
+    // TODO: Composite Query Filter will be needed to combine and add all filter to query in future
+    Query query = new Query("PublicRecipe");
+    query.setFilter(SearchUtils.getRecipeNameFilter(uQuery.getName()));
```
Aliblas, line 53: maybe

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+          datastoreRecipe.getKey().getId(),
+          (String) datastoreRecipe.getProperty("name"),
+          (String) datastoreRecipe.getProperty("imageBlobKey"),
+          (String) datastoreRecipe.getProperty("description")
+        )
+      );
+    }
+
+    // Sets the response type,converts the data into JSON, and sends the response
+    response.setContentType("JSON");
```
Aliblas: Is this meant to be application/json?

src/main/java/com/google/sps/util/SearchUtils.java
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Takes in an authorID and returns the query filter for the given author so it can be added to
+   * a query.
+   */
+  public static Query.FilterPredicate getAuthorFilter(Long authorID) {
+      return new Query.FilterPredicate("authorID", Query.FilterOperator.EQUAL, authorID);
+  }
+
+  /**
+   * Takes in an authorID and returns the query filter for the given author so it can be added to
```
Aliblas: Update comment

src/main/java/com/google/sps/util/SearchUtils.java
```diff
@@ -0,0 +1,98 @@
...
+   * a query.
+   */
+  public static Query.FilterPredicate getRecipeNameFilter(String recipeName) {
+      return new Query.FilterPredicate("name", Query.FilterOperator.EQUAL, recipeName);
+  }
```
Aliblas, line 79: Remember to account for standardized recipeName, versus non-standardized "name" property

Aliblas (approved):

src/main/java/com/google/sps/util/SearchUtils.java
Aliblas: eventually, anyway.

ez-may: How do you mean?

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+  private ArrayList<String> authors;
+  private String recipeName;
+
+  /**
+   * This constructor takes in a raw user query string, with special query symbols, and initializes
+   * instance variables using methods in SearchUtils to handle the query.
+   * TODO: Implement regex and make calls to initialize authors, and tagsAndIngredients using regex functions
+   */
+  public UserQuery(String query) {
+    String standardizedQuery= SearchUtils.standardize(query);
```
oturner24: nit: space after `standardizedQuery`

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+    this.tagsAndIngredients = new ArrayList<String>();
+    this.authors = new ArrayList<String>();
+    this.recipeName = SearchUtils.parseRecipeName(standardizedQuery);
+  }
+
+  /**
+   * Takes a new tag or ingeredient to search for and adds it to the tagsAndIngredients attribute.
+   * TODO: Implement ability to avoid duplicates
```
oturner24, line 43: Could you use a set to avoid duplicates?

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+   */
+  public void addTag(String tag) {
```
oturner24, line 45: Is there a reason you have `addTag` and not `addAuthor`?

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.sps.util.SearchUtils;
+import java.util.ArrayList;
+
+/**
+ * Represents the separated user query data which can be used to run different Datastore
+ * query requests.
+ */
+public class UserQuery {
+  private ArrayList<String> tagsAndIngredients;
+  private ArrayList<String> authors;
```
oturner24, line 26: Does it ever actually make sense to have multiple authors in the query? From what I understand from what you've discussed (correct me if I'm wrong), these attributes will be combined with the AND operator, ie if the user searches for `"#sweet #chocolate @olivia"` then you'll return recipes that have both tag=sweet **and** tag=chocolate **and** author=olivia. Since each recipe can only have one author, a query with multiple authors will always return empty results, right?

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+    // Creates the new Query and adds the filter to be used in Datastore
+    // The filter is currently only set to find an exact match name in Datastore
+    // TODO: Composite Query Filter will be needed to combine and add all filter to query in future
+    Query query = new Query("PublicRecipe");
+    query.setFilter(SearchUtils.getRecipeNameFilter(uQuery.getName()));
+
+    // Makes the query to the datastore and converts it to a list so it can operated on.
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery datastoreResult = datastore.prepare(query);
+    List<Entity> recipes = datastoreResult.asList(FetchOptions.Builder.withLimit(10));
```
oturner24: Store 10 as a constant with a descriptive name. Also, is there a reason the limit is 10?

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+
+    // Create a list of recipe entity information to be sent to the client, and then iterate through
+    // the datastore response to create the recipe objects and add them to the list
+    ArrayList<Recipe> clientRecipeInfo = new ArrayList<Recipe>();
+    for (int i = 0; i < recipes.size(); i++) {
+      Entity datastoreRecipe = recipes.get(i);
```
oturner24: if you're only using the index i to get a recipe, can you use a for each loop instead? ie:
`for (Entity datastoreRecipe : recipes)`

src/main/java/com/google/sps/servlets/Search.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Converts an object into JSON using Gson, but abstracts the need to make
+   * a Gson object in different parts of code.
+   **/
+  private String convertToJson(Object target) {
+    Gson gson = new Gson();
+    return gson.toJson(target);
+  }
```
oturner24: These seem like pretty generic methods. If they'd be useful for other classes as well, consider moving them to a utils file

src/main/java/com/google/sps/util/SearchUtils.java
```diff
@@ -0,0 +1,98 @@
...
+public class SearchUtils {
+  /**
+   * Takes a string and returns a standard version of the string depending on the implementation.
+   * This is used for processing a string before parsing for different query data.
+   * Current implementation returns strings trimmed and in lowercase.
+   */
+  public static String standardize(String str) {
+   String standard = str.trim();
+   standard = standard.toLowerCase();
+   return standard;
```
oturner24: could you simplify this to `return str.trim().toLowerCase();`?

src/main/java/com/google/sps/util/SearchUtils.java
```diff
@@ -0,0 +1,98 @@
...
+  public static String parseRecipeName(String str) {
+    return str;
+  }
+
+  /**
+   * Takes in an authorID and returns the query filter for the given author so it can be added to
+   * a query.
+   */
+  public static Query.FilterPredicate getAuthorFilter(Long authorID) {
+      return new Query.FilterPredicate("authorID", Query.FilterOperator.EQUAL, authorID);
```
oturner24: nit: it looks like the ident is off on this line?

src/main/java/com/google/sps/util/SearchUtils.java
```diff
@@ -0,0 +1,98 @@
...
+      return new Query.FilterPredicate("name", Query.FilterOperator.EQUAL, recipeName);
+  }
+
+  /*
+   * Takes a tag a returns a query filter which can be used to check if the tag is in the tag or
+   * ingredient list of the entity property.
+   */
+  public static Query.CompositeFilter getTagFilter(String tag) {
+      // Filter for whether the tag is in the ingredient list
+      Query.Filter ingredFilter = new Query.FilterPredicate("ingredients", Query.FilterOperator.IN, tag);
```
oturner24: nit: for `ingredFilter`, don't abbreviate variable names unless it's a well-known abbreviation.

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -1,3 +1,10 @@
+/**
+ * The function to load the main page and display content properly 
+ */
+mainPageLoad = () => {
+
```
oturner24: can you add either content here, or a TODO comment to describe what you intend to implement here?

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -57,15 +64,22 @@ createRecipeCard = (divID, recipeInfo) => {
...
 createElement = (htmlTag, object = "", tagOptions = {}) => {
-    // Tries to create an HTML element with the given tag
+    // Create an HTML element with the given tag
     let htmlElement = document.createElement(htmlTag.toLowerCase());
 
-    let htmlText = document.createTextNode(object);
-    htmlElement.appendChild(htmlText);
+    // If the object parameter is not undefined or an empty string, it 
+    // creates a textnode and adds it to the HTML element
+    if (object !== undefined || object !== "") {
```
oturner24: Should this be &&? Same for the if-statement below

oturner24 (changes requested):

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and fixed

src/main/java/com/google/sps/servlets/Search.java
ez-may: Ack and Fixed

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and fixed

src/main/java/com/google/sps/servlets/Search.java
ez-may: Ack and fixed

src/main/java/com/google/sps/servlets/Search.java
ez-may: Yes this was my bad, I fixed it

src/main/java/com/google/sps/util/SearchUtils.java
ez-may: Missed it in a ctrl+c ctrl+v moment, I fixed it

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and fixed

src/main/java/com/google/sps/data/UserQuery.java
ez-may: This is a good suggestion, and right off the top of my head, I would say yes, probably. I've read through some of the Javadoc's for the collection interface and subimplementing classes, so I think a set should work from what I remember reading and available methods. Since this isn't implemented yet, should I go ahead and make the change or can it wait for another PR?

oturner24: future PR is fine :)

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Originally we had thought about having an extended side menu where you could add more filters/tags to a search form, much like the one many search pages for online commerce have. Although I'm still not sure if it will be possible to implement this, in the hypothetical case we do, addtag() exists so after the tagsAndIngredients attribute is initialized based on the query string, and data from the extra form can be added in using this method (which would see a different implementation most likely) .

src/main/java/com/google/sps/data/UserQuery.java
ez-may: I don't think it does right now, and I did this more thinking about implementing author search to be explicitly an **OR** operator in the future. I was thinking the back end could handle these differently, and add some more meaningful search options for a user.

src/main/java/com/google/sps/servlets/Search.java
ez-may: Ok, I can do that.
I set 10 because the idea is to eventually limit results, implement pagination, and use datastore cursors to retrieve the recipe information. 10 is the limit I thought would be good for testing, but it's not a hard set number

src/main/java/com/google/sps/data/UserQuery.java
oturner24: Makes sense, thanks for the explanation

src/main/java/com/google/sps/servlets/Search.java
ez-may: This is a good point. I've been using them many other places, so this would eliminate the need to copy/paste

src/main/java/com/google/sps/data/UserQuery.java
oturner24: Okay, that sounds fine. If you are using AND for some elements of the search and OR for other parts, just be sure that's obvious from the frontend. Basically, the user shouldn't be surprised by how the search works.

src/main/java/com/google/sps/servlets/Search.java
oturner24: That sounds reasonable, thanks for explaining

src/main/java/com/google/sps/util/SearchUtils.java
ez-may: I could, but I'm not sure if any problems or changes will be made in the future that will require the standardizing method to be more comprehensive, and for that reason I decided to leave it split. I could simplify it to `String standard = str.trim().toLowerCase()` and then `return standard;` if you think that's better

src/main/java/com/google/sps/util/SearchUtils.java
Aliblas: If I understand correctly, recipeName is converted to all lower-case, trimmed, cleaned of special chars, etc. Is the same done for "name" property before they are compared?

src/main/java/com/google/sps/util/SearchUtils.java
oturner24: At this point the logic is pretty simple, so I don't think you'd run into serious issues if you needed to add more to this method in the future. What you suggested looks good to me, but it's up to you - I'd suggest choosing whichever one you think is the most readable.

src/main/java/com/google/sps/servlets/Search.java
ez-may: Ack and fixed. I don't know why I had originally thought this wasn't possible

src/main/java/com/google/sps/util/SearchUtils.java
ez-may: Ack and fixed

src/main/java/com/google/sps/util/SearchUtils.java
ez-may: Ack and fixed

src/main/webapp/javascript/recipe-rendering.js
ez-may: Ack and added TODO

src/main/webapp/javascript/recipe-rendering.js
ez-may: Ahh I see what I did here. For || I would have had to use ===, but for !== I have to use &&.

ez-may: Ack and Fixed

src/main/java/com/google/sps/util/SearchUtils.java
ez-may: Sounds good

src/main/java/com/google/sps/util/SearchUtils.java
ez-may: So I talked to isabel about this, and mentioned the basic cleaning process with her regarding tags and such, but I'll go over it with her to make sure we're on the same page.

src/main/java/com/google/sps/data/TestData.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class TestData {
+  /**
+   * Nested class used to initialize recipes for testing basic recipe name searching
+   */
+  public static final class RecipeNameSearch {
+    // Properties for test recipes in this static class
+    private static final ArrayList<String> TAG_LIST = new ArrayList<String>(
+        Arrays.asList("test", "name", "search"));
+    private static final ArrayList<String> INGRED_LIST = new ArrayList<String>(
```
oturner24: `INGRED_LIST` -> `INGREDIENT_LIST`
Don't use abbreviations in variable names unless they're commonly used.

src/main/java/com/google/sps/data/TestData.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final Entity RECIPE_009 = TestUtil.createRecipeEntity(
+      ENTITY_NAME, ALT_RECIPE_NAME, "THIS IS TEST RECIPE 009 FOR THE SEARCH BY NAME TEST",
+      TAG_LIST, INGRED_LIST, STEPS_LIST, -1);
+    private static final Entity RECIPE_010 = TestUtil.createRecipeEntity(
+      ENTITY_NAME, ALT_RECIPE_NAME, "THIS IS TEST RECIPE 010 FOR THE SEARCH BY NAME TEST",
+      TAG_LIST, INGRED_LIST, STEPS_LIST, -1);
+
+    // Array of all test entities
+    private static Entity[] testRecipes = new Entity[]{RECIPE_001, RECIPE_002, RECIPE_003, RECIPE_004,
+      RECIPE_005, RECIPE_006, RECIPE_007, RECIPE_008, RECIPE_009, RECIPE_010};
```
oturner24: Seems like it would be easier to create these programmatically in the for loop below. That would also let you pass in parameters to `initializeData` to customize your test data, for example the number of recipes to create.

src/main/java/com/google/sps/data/TestData.java
```diff
@@ -0,0 +1,74 @@
...
+    
+    // Creates the datastore and puts all the test recipes into the Datastore
+    public static void initializeData() {
+        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        for (int i = 0; i < 10; i++) {
+          datastore.put(testRecipes[i]);
+        }
+        return;
```
oturner24: You can remove this

src/main/java/com/google/sps/data/TestData.java
```diff
@@ -0,0 +1,74 @@
...
+   * Nested class used to initialize recipes for testing basic recipe name searching
+   */
+  public static final class RecipeNameSearch {
+    // Properties for test recipes in this static class
+    private static final ArrayList<String> TAG_LIST = new ArrayList<String>(
+        Arrays.asList("test", "name", "search"));
+    private static final ArrayList<String> INGRED_LIST = new ArrayList<String>(
+        Arrays.asList("none"));
+    private static final ArrayList<String> STEPS_LIST = new ArrayList<String>(
+        Arrays.asList("none"));
```
oturner24: Why are these not just empty?

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -0,0 +1,60 @@
...
+  public ArrayList<String> getTagsAndIngredients() {
+      return this.tagsAndIngredients;
+  }
+
+  public ArrayList<String> getAuthors() {
+      return this.authors;
+  }
+
+  public String getName() {
+      return this.recipeName;
```
oturner24: nit: the indentation for the return statements looks off (too far indented, I think)

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -0,0 +1,82 @@
...
+    // Create a list of recipe entity information to be sent to the client, and then iterate through
+    // the datastore response to create the recipe objects and add them to the list
+    ArrayList<Recipe> clientRecipeInfo = new ArrayList<Recipe>();
+    for (Entity datastoreRecipe : recipes) {
+      clientRecipeInfo.add(
+        new Recipe(
+          datastoreRecipe.getKey().getId(),
+          (String) datastoreRecipe.getProperty("name"),
+          (String) datastoreRecipe.getProperty("imageBlobKey"),
+          (String) datastoreRecipe.getProperty("description")
```
oturner24, line 72: Is it possible for these to be missing or null? Would it make sense to check that they exist before creating the recipe object?

src/main/java/com/google/sps/util/TestUtil.java
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.util;
+
+import com.google.appengine.api.datastore.Entity;
+import com.google.sps.data.Recipe;
+import java.util.ArrayList;
+
+public class TestUtil {
```
oturner24, line 21: Do you have future ideas for this class vs the TestData class? Since they're both small and do related things, it seems like you could combine them, unless you know that in the future it makes more sense for them to be separate.

oturner24 (approved):

src/main/java/com/google/sps/data/TestData.java
ez-may: Ack and Fixed

src/main/java/com/google/sps/data/TestData.java
ez-may: Ack and fixed;

src/main/java/com/google/sps/data/TestData.java
ez-may: I was uncertain what would happen if I didn't initialize these values in the backend, once the recipes were retrieved and sent to the front end. That's why I use a helper method to take all the required recipe data and then return an entity. I could have left it as an empty array, too, which is maybe what you're asking about, and I decided against that because if I could create very specific tags for the tests, when I went to delete the entities, I could use the tag search methods I have to remove the test recipes based on name + tag combinations.

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and fixed

src/main/java/com/google/sps/data/TestData.java
oturner24: Yeah, the second one (leaving it as an empty array) is what I was asking about, but if you have a reason to initialize it with "none" then that's fine. Maybe add a comment explaining why you're doing it this way (doesn't have to be part of this PR)

src/main/java/com/google/sps/servlets/SearchServlet.java
ez-may: It shouldn't be possible. I talked with Isabel about this earlier on, and I believe all of there a mandated fields for the user so this data should always be present.

src/main/java/com/google/sps/data/TestData.java
ez-may: Ack and fixed, though slightly different to how you proposed.

src/main/java/com/google/sps/util/TestUtil.java
ez-may: Yes, the idea is that as I begin working on more complex parts of search and want to make sure that my implementation is working and doesn't break, TestData will have static classes which represent a data set for a specific test. Each static class will have an initialize and clear data method for putting their data into the Datastore and deleting it during testing. The variables are public and static so they can be accessed from the test and values can be asserted using JUnit. 
Eventually, I hope to have one of these which can initialize data in the "real Datastore" for front end testing, and for the other nested classes have them take a parameter in the initialize and clear methods for a mock datastore reference which they will use.

TestUtil will have any functions that can be reused among the nested classes in TestData which help abstract functionality, but no actual Entity/recipe information.

ez-may:
Screenshot 2020-07-17 at 1 09 18 PM
This is a screen shot to show the result of the first call to initialize data in the data store, since there's no other good way to verify everything working yet.

src/main/java/com/google/sps/data/TestData.java
```diff
@@ -0,0 +1,70 @@
...
+              for (Entity recipeElement : tempEntityList) {
+                  datastore.put(recipeElement);
+              }
+          }
+        } catch (ArrayIndexOutOfBoundsException AIOB) {
+            Utils.SOP(String.join("\n",
+              "An array out of bounds exception was just caught.",
+              "Please check that the size of TEST_SET_NAMES and TEST_SET_SIZES are equal to NUM_SETS"
+            ));
+        }
```
oturner24: Instead of catching an exception, can you check that they all have equal lengths before the for loop? And then you can print the same message and return early.

src/main/java/com/google/sps/util/TestUtil.java
oturner24: Sounds good, thanks for the explanation.

oturner24 (approved):
Still LGTM!

PR #17, opened 2020-07-13
User class

Aliblas:
This includes a class instance initialization method getTestUser which creates a single universal instance of a user for testing purposes.
There is a lot of code commented out which I will probably move into a different branch, but I wanted to get a PR open for now.


src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+import com.google.api.client.json.jackson2.JacksonFactory;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.*;
```
ez-may: You should remove the wildcard import per Java style guide:
https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+    "picture": "https://lh4.googleusercontent.com/-kYgzyAWpZzJ/ABCDEFGHI/AAAJKLMNOP/tIXL9Ir44LE/s99-c/photo.jpg",
+    "given_name": "Test",
+    "family_name": "User",
+    "locale": "en"
+  */
+
+  /**
+   * Prevents usage of default constructor
+   */
+  private User() { }
```
ez-may: I may be mistaken, but I believe you shouldn't need this since you already define a constructor below. Checkout this SO too:
https://stackoverflow.com/a/4488766

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+   * This test method will not be removed from the webapp prior to deployment.
+   */
+  public static User getTestUser() {
+    User testUser = new User();
+    String testID = "000000000000000000000";
+    Key userKey = KeyFactory.createKey("User", testID);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    try {
+        testUser.entity = datastore.get(userKey);
+    } catch(Exception e) {
```
ez-may: The style guide isn't clear here and I'm not totally sure, but if I remember correctly, I believe it's best practice to not use a general exception in Java, but rather to catch specific errors and handle them . It looks like you already know the name of the exception, so including it could be a small change.

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+        testUser.entity.setProperty("cookbook", emptyJsonArray);
+        testUser.entity.setProperty("userRecipes", emptyJsonArray);
+        testUser.entity.setProperty("userRecipes", emptyJsonArray);
+        testUser.entity.setProperty("planner", emptyJsonArray);
+    }
+    return testUser;
+  }
+
+  public long getDisplayName() {
+    return (long)entity.getProperty("displayName");
```
ez-may: Nit: space between cast parenthesis and expression

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+      list.add(id);
+    }
+    entity.setProperty(recipeListType, gson.toJson(list));
+    put();
+  }
+
+  /**
+   * Creates a datastore object and stores current entity in database.
+   */
+  private void put() {
```
ez-may: To me, this name seems mildly confusing since (in my experience with it) this function keyword is commonly used as obj.put(something) where "something" is then put in "obj." When you call this method above, I almost left a comment because I didn't know what you were calling and seemed like a typo. Is there a clearer name you could use, even if it's putEntity() ?

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+  }
+
+  /**
+   * Gets the (String) Json array of a given recipe list.
+   * @Param property Property name of entity list to be obtained.
+   * Returns Parsed array as ArrayList<Long>
+   */
+  private ArrayList<Long> getPropertyArrayList(String property) {
+    Gson gson = new Gson();
+    String jsonArray = (String)entity.getProperty(property);
```
ez-may: Nit: space between casting and expression

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+    Key userKey = KeyFactory.createKey("User", testID);
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    try {
+        testUser.entity = datastore.get(userKey);
+    } catch(Exception e) {
+        // On EntityNotFoundException, create initial instance.
+        testUser.entity = new Entity(userKey);
+        String emptyJsonArray = "[ ]";
+        testUser.entity.setProperty("displayName", "Test Person");
+        testUser.entity.setProperty("cookbook", emptyJsonArray);
```
ez-may: I'm not sure what the scope of your testing will be, but if you set all these properties to point to the same string, it may be the case that they will all end up pointing to the same reference, and a change in one of the properties would be reflected in all of them. Additionally, if your test user is storing string data in entity properties which are expected to have ArrayList<String> data everywhere else, you may run in to unexpected behavior/errors as you try to use this code later on to test different parts of the website.

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
-    this.id = id;
-    this.email = email;
-    this.username = username;
-    planner = new ArrayList<Long>();
-    cookbook = new ArrayList<Long>();
-    userRecipes = new ArrayList<Long>();
+  /**
+    // User Entity Properties:
+
+    "displayName" : "Example Name"
```
ez-may: Nit: I know this is all commented out, but in some instances you have a space between the key, colon, and value, and in other cases you have the key and colon together, and a space between the colon and key value. The style guide doesn't seem to specify this, so I would just say stay consistent.

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,195 @@
...
+    removeIDFromRecipeList(id, "cookbook");
   }
 
   public void removeRecipeFromUserRecipes(long id) {
-    userRecipes.remove(new Long(id));
+    removeIDFromRecipeList(id, "userRecipes");
+  }
+
+  /**
+   * @Param id recipe ID to be removed from current User entity.
```
ez-may, line 139: I think we're supposed to still include a brief description of the methods even if the name is pretty obvious

ez-may (changes requested):

src/main/java/com/google/sps/data/User.java
Aliblas: yes putEntity will do well

src/main/java/com/google/sps/data/User.java
Aliblas: This is true

src/main/java/com/google/sps/data/User.java
Aliblas: The entity is a datastore Entity,  Entity.setProperty copies the string literal "[ ]" and does not hold any reference.
Also, the String Json arrays are fully in abstraction, the get and add methods control conversion to and from ArrayList data, and no one will handle the data without them.
I have tested the functionality of the methods using a local test Servlet.

src/main/java/com/google/sps/data/User.java
Aliblas: https://google.github.io/styleguide/javaguide.html#s7.3.1-javadoc-exception-self-explanatory
So not generally the case, but I thought these two could use a tidbit.

src/main/java/com/google/sps/data/User.java
Aliblas: I left it there for the method initializing the test user, but I changed the comment to reflect this.

src/main/java/com/google/sps/data/User.java
ez-may: SGTM then!

src/main/java/com/google/sps/data/User.java
ez-may: Sounds Good

ez-may (approved):
LGTM

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,198 @@
...
-  private ArrayList<Long> planner;
-  private ArrayList<Long> userRecipes;
-  private String email;
-  private String id;
-  private String username;
-  // shopping list will not be implented until after MVP
-  private ArrayList<String> shoppingList;
+  // API variables:
+  // client ID is generated from Ali's google APIs credentials page.
+  private static String CLIENT_ID = "1034390229233-u07o0iaas2oql8l4jhe7fevpfsbrtv7n.apps.googleusercontent.com";
```
oturner24: This should probably be private static final

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,198 @@
...
+    // "email" OAuth scopes to the application.
+    // "sub" or Subject refers to the user-unique ID number.
+    "sub" : "110169484474386276334"
+    "email" : "testuser@gmail.com",
+    "email_verified" : "true",
+    "name" : "Test User",
+    "picture" : "https://lh4.googleusercontent.com/-kYgzyAWpZzJ/ABCDEFGHI/AAAJKLMNOP/tIXL9Ir44LE/s99-c/photo.jpg",
+    "given_name" : "Test",
+    "family_name" : "User",
+  */
```
oturner24: A couple things about this comment block:
1) Since you have a comment block with `/* ... */` you can remove all the`//` within the block
2) It seems like it would be more helpful to put the comments above the variables they relate to, like:
`/* comment about payload */`
`private Payload payload;`
`/* comment about entity */`
`private Entity entity;`
3) Are the long IDs stored as strings? It seems confusing to be representing them as "longId1", etc when you would be expecting a number, ie 1 or "1" (depending on the answer to the first question)

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,198 @@
...
+   * Used for initialization with static methods.
+   */
+  private User() { }
+
+  /**
+   * Constructs a User instance based on a unique id_token.
+   * Verify user using google's ID token verifyer API.
+   * This method is commented out because it does not work yet.
+   */
+  public User(String idTokenString) {
```
oturner24: In general it's probably best to just not include constructors/methods that don't work in PRs, but if it doesn't work and it's not used anywhere, you should comment out the whole thing, not just the body of it.

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,198 @@
...
+    // }
+
+    // payload = idToken.getPayload();
+  }
+
+  /**
+   * This creates a standard universal user instance regardless of login status.
+   * This test method will not be removed from the webapp prior to deployment.
+   */
+  public static User getTestUser() {
```
oturner24: It seems a little weird to return an instance of the object by a static method..can you make this a private helper method wherever you need to use it? Alternatively, you could move this logic to the constructor for now? Either use the constructor with no parameters, or have a boolean parameter indicating whether it's a test user or not.

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,198 @@
...
+    return (long) entity.getProperty("displayName");
+  }
+
+  public void setDisplayName(String name) {
+    entity.setProperty("displayName", name);
+    putEntity();
+  }
+
+// Cookbook and Planner currently support only one type of Recipe entity (namely public).
+// Likewise, userRecipes should only include private recipes.
```
oturner24: nit: use `/* */` for multi line comments

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,198 @@
...
+   * Helper method for adding recipe IDs to specific user recipe list.
+   * @Param id recipe ID to be added to current User entity.
+   * @Param recipeListType Property name of the list to add to.
+   */
+  private void addIDToRecipeList(long id, String recipeListType) {
+    Gson gson = new Gson();
+    ArrayList<Long> list = getPropertyArrayList(recipeListType);
+    if (!list.contains(id)) {
+      list.add(id);
+    }
```
oturner24, line 160: Have you considered using a hash set? You could avoid doing this check, and it has the added benefit of a constant time remove in the above method. Using a hash set doesn't guarantee an iteration order though, if that is something you're dependent on.

src/main/java/com/google/sps/servlets/UserActionServlet.java
```diff
@@ -14,35 +14,37 @@
...
+import com.google.sps.data.User;
 import java.io.IOException;
-import java.io.PrintWriter;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+/**
+ * provides get method for adding a recipe ID to one of the three User recipe lists.
```
oturner24, line 26: super nit: capitalize the beginning of the comment

src/main/java/com/google/sps/servlets/UserActionServlet.java
```diff
@@ -14,35 +14,37 @@
...
-@WebServlet("/user-status")
-public class UserStatusServlet extends HttpServlet {
-
-  // Response: returns some user information JSON from the current user's sign-in API profile.
-  // Data: "isUserLoggedIn" : boolean
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    response.setContentType("application/json");
+    String destination = (String) request.getParameter("destination");
+    long ID = Long.parse((String) request.getParameter("inputID"));
```
oturner24: nit: lowercase id

src/main/java/com/google/sps/servlets/UserActionServlet.java
```diff
@@ -14,35 +14,37 @@
...
-import java.io.PrintWriter;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+/**
+ * provides get method for adding a recipe ID to one of the three User recipe lists.
+ * request parameters: "destination", "inputID"
+ * "destination" may be one of three user recipe list: "cookbook", "userRecipes", "planner".
```
oturner24, line 28: It seems error prone to depend on the request having the exact string you expect, could you use an enum instead?

src/main/java/com/google/sps/servlets/UserActionServlet.java
```diff
@@ -14,35 +14,37 @@
...
+    
+    // Must be changed to implement distinct users with sign-in API.
+    User user = User.getTestUser();
 
-    JsonObject userInfo = new JsonObject();
-    userInfo.addProperty("isUserLoggedIn", UserHelper.isUserLoggedIn());
-    if (UserHelper.isUserLoggedIn()) {
-        System.out.println("User is logged in");
-        // other properties may be added in the future.
+    if (destination == "cookbook") {
```
oturner24: Use .equals() to compare strings. Here's a [stackoverflow](https://stackoverflow.com/questions/7520432/what-is-the-difference-between-and-equals-in-java#:~:text=The%20equals()%20method%20compares,same%20String%20instance%20or%20not.&text=On%20the%20other%20hand%2C%20the,to%20the%20same%20String%20instance), the tl;dr is that .equals() compares the values, while == compares the object references

oturner24 (changes requested):

src/main/java/com/google/sps/data/User.java
Aliblas: I will definitely consider this for a future pr, I'll probably revise for efficiency once I at least have the user verification working properly.

src/main/java/com/google/sps/servlets/UserActionServlet.java
Aliblas: on that note, which would you say is more correct: fooID or fooId?

src/main/java/com/google/sps/servlets/UserActionServlet.java
Aliblas: I agree with this but I am not sure how I would enforce it from the user's end. How would I make the enum options available to them from a fetch request?

src/main/java/com/google/sps/servlets/UserActionServlet.java
oturner24: I think fooId is more in line with the style guide, since we usually camel case abbreviations, for example: someUrl, someGps, etc

src/main/java/com/google/sps/servlets/UserActionServlet.java
oturner24: Good question, to be honest I'm not totally sure. I did some quick googling, and you could try using `@JsType` like described [here](http://www.luigibifulco.it/blog/en/blog/gwt-2-8-0-jsinterop), or try asking the VSEs. 

That being said, I wouldn't spend too much time on this - it's probably also fine to make these strings constants, and add a comment on the front end referencing the constants in this file.

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,188 @@
 package com.google.sps.data;
 
+// import com.google.api.client.extensions.appengine.http.UrlFetchTransport;
```
oturner24: remove this?

src/main/java/com/google/sps/data/User.java
```diff
@@ -1,57 +1,188 @@
...
+      entity.setProperty("planner", emptyJsonArray);
+    }
+  }
+
+  /**
+   * Returns the String user unique ID.
+   * Currently returns ID of test user instance.
+   */
+  String getID() {
+    return "000000000000000000000";
```
oturner24: Maybe store this in a constant to be referenced here and in the constructor?

src/main/java/com/google/sps/data/User.java
oturner24: Sure that sounds good

src/main/java/com/google/sps/servlets/UserActionServlet.java
```diff
@@ -14,35 +14,37 @@
...
-    if (UserHelper.isUserLoggedIn()) {
-        System.out.println("User is logged in");
-        // other properties may be added in the future.
+    if (destination.equals("cookbook")) {
+      user.addRecipeToCookbook(id);
+    } else if (destination.equals("userRecipes")) {
+      user.addRecipeToUserRecipes(id);
+    } else if (destination.equals("planner")) {
+      user.addRecipeToPlanner(id);
     }
```
oturner24: It's probably also a good idea to have an else that will return an error message if the destination isn't one of the values you expect.

oturner24 (approved):

src/main/java/com/google/sps/servlets/UserActionServlet.java
Aliblas: Thanks, I I will go with the second solution for now, but perhaps I'll look into @JsType in a future pr

PR #20, opened 2020-07-16
Edit recipe

isamitre:
EditRecipeServlet pulls the stored recipe from datastore using a key made from the recipe's id
it compares the form's values to the stored values and if any of the form's values are empty, the stored value will be the one saved
it's all in a huge try/catch bc the compiler was getting mad at me otherwise; if there's a better way of catching the exception, let me know


isamitre:
image

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
```diff
@@ -33,7 +33,17 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
-    String uploadUrl = blobstoreService.createUploadUrl("/new-recipe");
+    String divID = request.getParameter("divID");
+    String uploadUrl = "";
+    if (divID.equals("new-recipe-modal")) {
+      uploadUrl = blobstoreService.createUploadUrl("/new-recipe");
+        System.out.println("     BlobstoreUploadUrlServlet: /new-recipe");
```
ez-may: Did you mean to keep the print statement? If so, the indentation here is also wrong.
Same for the print below

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+
+    // declaring the lists
+    ArrayList<String> tags;
+    ArrayList<String> ingredients;
+    ArrayList<String> steps;
+
+    long id = Long.parseLong(idResponse);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
```
ez-may, line 36: Nit: remove extra spaces

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+    ArrayList<String> tags;
+    ArrayList<String> ingredients;
+    ArrayList<String> steps;
+
+    long id = Long.parseLong(idResponse);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    
+    Key key = KeyFactory.createKey("PrivateRecipe", id);
+    try {
```
ez-may, line 40: You have some similar code from new recipe if I remember correctly (calling out the tag separation using streams). Why not put this functionality into a method, maybe even a utility file, and call it from there?

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+      Entity recipeEntity = datastore.get(key);
+
+      // if the name input in the form is empty, the original value will be saved
+      if (name.equals("")) {
+        name = (String) recipeEntity.getProperty("name");
+      } else {
+        // formats the form response
+        name.trim();
```
ez-may: I don't think you're getting the results  you expect. I believe you need to do `name = name.trim()`
Ref: https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#trim--

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+
+      // if the ingredients input in the form is empty, the original value will be saved
+      if (stepsResponse.equals("")) {
+        steps = (ArrayList<String>) recipeEntity.getProperty("tags");
+      } else {
+        steps = new ArrayList<String>(Arrays.asList(Arrays.stream(stepsResponse.split("\n")).map(String::trim).toArray(String[]::new)));
+        steps.removeAll(Arrays.asList("", null, "\n", "\r\n", "\r"));
+      }
+
+      recipeEntity.setProperty("name", name);
```
ez-may, line 71: This is another area where I could see some abstraction being helpful. A utility function which could take in a recipe object and a Datastore reference (or even makes it internally), and then create an entity and put it in the Datastore could cleanup a bit here and in new recipe, since the code looks very similar.

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+      recipeEntity.setProperty("tags", tags);
+      recipeEntity.setProperty("description", description);
+      recipeEntity.setProperty("ingredients", ingredients);
+      recipeEntity.setProperty("steps", steps);
+      recipeEntity.setProperty("imageBlobKey", (String) recipeEntity.getProperty("imageBlobKey"));
+
+      datastore.put(recipeEntity);
+
+    } catch (EntityNotFoundException e) {
+        System.out.println("entity not found exception");
```
ez-may: Is there a plan in the future to implement some different behavior to the catch block? I'm not familiar with what cases this catch block could be entered, but maybe including a TODO with the future implementation or comment explaining what happens when it's could help.

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
```
ez-may: Nit: remove extra space

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+      recipeEntity.setProperty("name", name);
+      recipeEntity.setProperty("tags", tags);
+      recipeEntity.setProperty("description", description);
+      recipeEntity.setProperty("ingredients", ingredients);
+      recipeEntity.setProperty("steps", steps);
+      recipeEntity.setProperty("imageBlobKey", (String) recipeEntity.getProperty("imageBlobKey"));
+
+      datastore.put(recipeEntity);
+
```
ez-may: Nit: remove extra space

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      // if the ingredients input in the form is empty, the original value will be saved
+      if (stepsResponse.equals("")) {
+        steps = (ArrayList<String>) recipeEntity.getProperty("tags");
+      } else {
+        steps = new ArrayList<String>(Arrays.asList(Arrays.stream(stepsResponse.split("\n")).map(String::trim).toArray(String[]::new)));
+        steps.removeAll(Arrays.asList("", null, "\n", "\r\n", "\r"));
+      }
+
```
ez-may, line 70: Nit: extra space

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,120 @@
...
+    long id = Long.parseLong(idResponse);
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    
+    Key key = KeyFactory.createKey("PrivateRecipe", id);
+    try {
+      Entity recipeEntity = datastore.get(key);
+
+      // if the name input in the form is empty, the original value will be saved
+      if (name.equals("")) {
```
ez-may: What if you initialized all the variables to be the data from Datastore, and then only have the if-statements to change the values if an empty string was not found? This could help shorten the code and remove some "else" blocks.

src/main/webapp/javascript/user-page.js
```diff
@@ -64,6 +65,70 @@ function addDeleteFunctionality(recipes){
...
+// tells the server to delete the recipe.
+// input is a recipe js object
+function deleteRecipe(recipe) {
+  const params = new URLSearchParams();
+  params.append("id", recipe.id);
+  fetch("/delete-recipe", {method: "POST", body: params});
+}
+
+// adds edit functionality to the edit button in the recipe cards
+function addEditFunctionality(recipes){
```
ez-may: Nit: space between bracket and parenthesis

src/main/webapp/javascript/user-page.js
```diff
@@ -73,33 +138,27 @@ setUpDivWithNoRecipes = (divID, message) => {
   recipesDiv.style.height = "100px";
 }
 
-// opens the recipe form modal
-function openModal() {
-  document.getElementById("recipe-modal").style.display = "block";
-  fetchBlobstoreUrl();
-}
-
 // not quite sure what this does; something with blobs
```
ez-may: If you have a better grasp of this method now that you've edited it some, add a more meaningful comment. If it's still some API or SO magic, it can stay.

src/main/webapp/pages/UserPage.jsp
```diff
@@ -67,13 +67,45 @@ <h1 class="center">Create Recipe</h1>
             <input type="file"  accept="image/*" name="image" id="image">
             <br><br>
             <div id="buttons">
-              <button class="modal-btn" id="cancel-btn" onclick="closeModal()">Close</button>
+              <button class="modal-btn" id="cancel-btn" onclick="closeModal('new-recipe-modal')">Cacnel</button>
```
ez-may: Nit: typo on cancel

src/main/webapp/javascript/user-page.js
```diff
@@ -73,33 +138,27 @@ setUpDivWithNoRecipes = (divID, message) => {
...
       .then((response) => {
         return response.text();
       })
       .then((imageUploadUrl) => {
-        const messageForm = document.getElementById('recipe-form');
-        messageForm.action = imageUploadUrl;
+        let form;
+        if (id == "new-recipe-modal") {
+          form = document.getElementById("new-recipe-form");
+        } else if (id == "edit-recipe-modal") {
```
ez-may, line 151: This is just a general comment, but I've noticed that a lot of your equality operators use ==, and I know that JS has different behavior for === and ==. If you're aware of that and are sure about your decisions, then that's fine, but I wanted to point it out because it can cause some pretty weird situations.

ez-may (changes requested):

src/main/java/com/google/sps/servlets/BlobstoreUploadUrlServlet.java
isamitre: yeah it was just so I could see it when debugging

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: in the new recipe servlet, I do some testing to see if the imageblobkey is viable and I set it in the if statements if it is whereas in the edit recipe servlet I do it all at once so if I wanted to abstract it I'd have to make another constructor for a recipe that does not include an imageblobkey and I don't think it's worth it; I think it's fine as is

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: tbh I don't see a scenario where the key will not point to an entity bc I get the id I'm using to make the key from the datastore which goes to the json which is sent to the frontend in a hidden input the user cannot see or modify and then back to the backend
I only put the try catch because the compiler would get mad at me otherwise
there is no future implementation

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
ez-may: Ok, sounds good then

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
ez-may: Ok, maybe a small comment here would help then documenting that this shouldn't be entered under normal circumstance based on what the user does. That way, if someone ever got the log statement they know not to look there.

ez-may (approved):
LGTM! Thanks for the changes

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: okay cool will do

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,87 @@
...
+      ArrayList<String> tags = (ArrayList<String>) recipeEntity.getProperty("tags");
+      String description = (String) recipeEntity.getProperty("description");
+      ArrayList<String> ingredients = (ArrayList<String>) recipeEntity.getProperty("ingredients");
+      ArrayList<String> steps = (ArrayList<String>) recipeEntity.getProperty("steps");
+
+      // if the name input in the form is not empty, the form value will be saved
+      if (!nameResponse.equals("")) {
+        name = nameResponse;
+        name = name.trim();
+      }
```
oturner24: You probably want to trim the nameResponse before you check if it's empty (what if the user puts a couple spaces?), and I think you can simplify the body of the if-statement: 
```
if (!nameResponse.trim().isEmpty()) {
    name = nameResponse.trim();
}
```

Same thing applies for the if-statements below, and the body of the description if-statement.

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -0,0 +1,87 @@
...
+      recipeEntity.setProperty("tags", tags);
+      recipeEntity.setProperty("description", description);
+      recipeEntity.setProperty("ingredients", ingredients);
+      recipeEntity.setProperty("steps", steps);
+      recipeEntity.setProperty("imageBlobKey", (String) recipeEntity.getProperty("imageBlobKey"));
+
+      datastore.put(recipeEntity);
+    } catch (EntityNotFoundException e) {
+      // in normal circumstances, this won't happen bc user has not access to id
+      System.out.println("entity not found exception");
```
oturner24, line 81: nit: add a little more detail to this error message so it's obvious where it's coming from / what's happening

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -37,30 +38,26 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     // redirects user back to UserPage
     // TODO: inform the user that they're missing stuff
     if (name.equals("") || tagsResponse.equals("") || description.equals("") || ingredientsResponse.equals("") || stepsResponse.equals("")){
```
oturner24, line 41: I know this isn't technically part of this PR, but you probably also want to trim the responses before checking that they're empty here, ie: 
```
if (name.trim().isEmpty() || tagsResponse.trim().isEmpty() .... )

oturner24 (approved):

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: that makes sense will do! though I'm going to put the trim to lines 44-48 when I'm getting the form responses

PR #21, opened 2020-07-17
User class refactor

Aliblas:
User class uses handles recipe keys now. SaveRecipeServlet has not been tested yet.


src/main/java/com/google/sps/data/User.java
```diff
@@ -11,14 +11,15 @@
...
+
 import java.io.IOException;
-import java.lang.reflect.Type;
+
 import java.lang.SecurityException;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+
```
isamitre: there shouldn't be whitespaces in the import statements
[Java Style Guide](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing)

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
```diff
@@ -24,41 +26,43 @@
...
-    User user;
-    try {
-      user = new User(idToken);
-    } catch(SecurityException e) {
-      System.out.println("ERROR: User verification failed.");
-      return;
-    }
+    String kind = (String) request.getParameter("kind");
+    long recipeId = Long.parseLong(request.getParameter("recipeId"));
+
```
isamitre: nit: remove extra whitespace

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
```diff
@@ -24,41 +26,43 @@
...
 
+    User currentUser = new User(userToken);
+
+    Key recipeKey;
+    
+    recipeKey = KeyFactory.createKey(kind, recipeId);
```
isamitre: declaration and initialization should be together

isamitre (approved):

src/main/java/com/google/sps/data/User.java
```diff
@@ -100,106 +97,92 @@ public String getId() {
...
+   * Helper method for retrieving ArrayList<Key> of given recipe list.
    */
-  private void addIdToRecipeList(long id, String recipeListType) {
-    Gson gson = new Gson();
-    ArrayList<Long> list = getPropertyArrayList(recipeListType);
-    if (!list.contains(id)) {
-      list.add(id);
+  @SuppressWarnings("unchecked") // Compiler cannot verify generic property type.
+  private ArrayList<Key> getRecipeList(String listName) {
+    ArrayList<Key> arrayListValue = (ArrayList<Key>) entity.getProperty(listName);
```
AlexanderEllis, line 156: Is there a better name for what this is? Maybe recipeListValue or something that indicates it's a list of recipe keys?

src/main/java/com/google/sps/data/User.java
```diff
@@ -76,20 +77,16 @@ public User(String idTokenString) throws SecurityException {
...
-      // On EntityNotFoundException, create initial instance.
+      // Create initial User instance, default name from idToken payload.
       entity = new Entity(userKey);
-      String emptyJsonArray = "[ ]";
-      // displayName defaults to given name from Gmail account.
-      entity.setProperty("displayName", (String) payload.get("name"));
-      entity.setProperty("cookbook", emptyJsonArray);
-      entity.setProperty("userRecipes", emptyJsonArray);
-      entity.setProperty("planner", emptyJsonArray);
+      entity.setProperty("name", (String) payload.get("name"));
```
AlexanderEllis: Would you want to use setName here instead, which would also call uploadEntity?

src/main/java/com/google/sps/data/User.java
```diff
@@ -76,20 +77,16 @@ public User(String idTokenString) throws SecurityException {
     }
 
     payload = idToken.getPayload();
-
+    datastore = DatastoreServiceFactory.getDatastoreService();
     Key userKey = KeyFactory.createKey("User", getId());
-    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
     try {
+      // Check if User already has a profile
```
AlexanderEllis: Suuuper nit: "Checks" and period at end of comment.

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+ * "recipeId" = Long
+ *  userToken is the user's Google id_token provided by user-auth.js/getIdToken().
+ * "userToken" = String
+ */
+@WebServlet("/save-recipe")
+public class SaveRecipeServlet extends HttpServlet {
+
+  private static final String COOKBOOK = "cookbook";
+  private static final String PLANNER = "planner";
+  private static final String PUBLIC_KIND = "PublicRecipe";
```
AlexanderEllis: Do you want to just have this be PUBLIC_RECIPE instead of PUBLIC_KIND? Same for private below.

Side note but is it possible to have these in a common file that both this and User.java import so you don't have to define COOKBOOK and PLANNER in two places? If not that's OK but it would be cleaner in the future if you ever had to change/add any.

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userToken = (String) request.getParameter("idToken");
+    String listName = (String) request.getParameter("listName");
+    Boolean isPublic = (Boolean) request.getParameter("isPublic");
+    long recipeId = Long.parseLong(request.getParameter("recipeId"));
+
+    User currentUser = new User(userToken);
+
+    String recipeKind = (isPublic) ? recipeKind = PUBLIC_KIND : recipeKind = PRIVATE_KIND;
```
AlexanderEllis: Looks like you may be setting recipeKind twice here inadvertently, once in each of the ternary options and once at the end. Does the following work instead?

`String recipeKind = (isPublic) ? PUBLIC_KIND : PRIVATE_KIND;`

src/main/java/com/google/sps/data/User.java
```diff
@@ -27,6 +24,10 @@
   private static final String CLIENT_ID = "1034390229233-u07o0iaas2oql8l4jhe7fevpfsbrtv7n.apps.googleusercontent.com";
   // Payload class contains decrypted user information.
 
+  private static final String COOKBOOK = "cookbook";
```
AlexanderEllis: https://www.youtube.com/watch?v=dk01eeKMD_I

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
```diff
@@ -0,0 +1,65 @@
...
+    
+    Key recipeKey = KeyFactory.createKey(recipeKind, recipeId);
+    
+    if (listName.equals(COOKBOOK)) {
+      currentUser.addCookbookKey(recipeKey);
+    } else if (listName.equals(PLANNER)) {
+      currentUser.addPlannerKey(recipeKey);
+    } else {
+      System.out.println("Invalid destination " + destination + ". Recipe not added.");
```
AlexanderEllis: I think instead of destination you mean `listName`.

(side note if you wanted to add this file in a separate PR after you've tested it, that would be OK too)

AlexanderEllis (commented):
LGTM, a few comments to address. 

Did you want to add SaveRecipeServlet.java in a separate PR after you get a chance to test it? May be missing something but wanted to check.

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
Aliblas: I've decided to just write the literals instead, I think it makes more sense and is less redundant.

src/main/java/com/google/sps/data/User.java
Aliblas: ahahaha good one.

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
Aliblas: Yes I'll do that.

src/main/java/com/google/sps/servlets/SaveRecipeServlet.java
Aliblas: Yes I forgot ternary works apparently.

AlexanderEllis (approved):
Thanks for making those changes!

PR #24, opened 2020-07-20
Publish edit recipe

isamitre:
if a recipe is unpublished, then in edit recipe there is an option to publish the recipe
once published, the public version of the recipe cannot be edited or deleted (at least right now) even if the private version of the recipe is edited or deleted; for all intents and purposes they are now two different recipes

### DIFF
--- a/src/main/java/com/google/sps/util/FormHelper.java
+++ b/src/main/java/com/google/sps/util/FormHelper.java
@@ -26,7 +26,7 @@ public class FormHelper {
 
   // input: two Entities
   // deep copies the second from the first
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/sps/util/FormHelper.java
```diff
@@ -26,12 +26,13 @@
 
   // input: two Entities
   // deep copies the second from the first
-  public static void copyRecipeEntity(Entity first, Entity second) {
+  public static void copyFirstRecipeEntityToSecond(Entity first, Entity second) {
```
isamitre, line 29: if you can think of a better name, please let me know lol

PR #25, opened 2020-07-20
Edit both recipes

isamitre:
there's the functionality we discussed
editing a private recipe with a public counterpart edits both the private and public recipes
making a public recipe private deletes the public recipe
the delete button on the recipe card also deletes the public recipe if there is one


src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -48,6 +51,10 @@ public long getId() {
     return id;
   }
 
+  public void publish() {
```
oturner24: Do you also need an unpublish?

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -16,11 +25,36 @@
...
+    long privateRecipeID = Long.parseLong(request.getParameter("id"));
+    String name = request.getParameter("name");
+    String description = request.getParameter("description");
 
-    // PrivateRecipes are the user uploaded recipes so only PrivateRecipes can be deleted (right now)
-    Key recipeEntitykey = KeyFactory.createKey("PrivateRecipe", id);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.delete(recipeEntitykey);
+
+    /// delete the private recipe
```
oturner24: nit: only 2 //

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -16,11 +25,36 @@
...
+    Key privateKey = KeyFactory.createKey("PrivateRecipe", privateRecipeID);
+    datastore.delete(privateKey);
+
+    // make filters to find the public recipe
+    Filter nameFilter = new FilterPredicate("name", FilterOperator.EQUAL, name);
+    Filter descriptionFilter = new FilterPredicate("description", FilterOperator.EQUAL, description);
+    // TODO: further filter by author ID
+    Filter composFilter = CompositeFilterOperator.and(nameFilter, descriptionFilter);
```
oturner24: It seems like it's possible for this to be broken; what if the same author writes multiple recipes with the same name and description? This probably wouldn't be common, but it should break your system either.

Would it be possible to store a mapping of private recipe ID -> public recipe ID to make this deletion super easy? It probably makes sense to store this when you are publishing the recipe.

Or, if you're making the assumption that authors can't have duplicate recipe name + description, you should be enforcing that when the user creates a new recipe.

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -67,17 +78,65 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+        datastore.put(publicRecipeEntity);
+      }
+
+      // recipe was public
+      if (published) {
+        // make filters to the public recipe
+        Filter nameFilter = new FilterPredicate("name", FilterOperator.EQUAL, savedName);
+        Filter descriptionFilter = new FilterPredicate("description", FilterOperator.EQUAL, savedDescription);
+        // TODO: further filter by author ID
+        Filter composFilter = CompositeFilterOperator.and(nameFilter, descriptionFilter);
```
oturner24: same comment here as above, about either mapping private and public ideas, or enforcing this assumption when a user creates a recipe

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -16,11 +25,36 @@
...
-    Key recipeEntitykey = KeyFactory.createKey("PrivateRecipe", id);
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    datastore.delete(recipeEntitykey);
+
+    /// delete the private recipe
+    Key privateKey = KeyFactory.createKey("PrivateRecipe", privateRecipeID);
+    datastore.delete(privateKey);
+
+    // make filters to find the public recipe
+    Filter nameFilter = new FilterPredicate("name", FilterOperator.EQUAL, name);
```
ez-may: Why is this implementation preferred? Why not have the id of the public recipe saved as an attribute of the private recipe, this way you can just delete the recipe directly?

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
ez-may: Nit: Don't use abbreviated variable names.

src/main/webapp/pages/UserPage.jsp
```diff
@@ -92,16 +92,17 @@
...
-            <option value="public">public</option>
-            </select>
-            <br><br> -->
+            <div id="publishDiv">
+            <label for="privacy">Choose a privacy setting:</label>
+              <select id="privacy" name="privacy">
+                <option id="first"></option>
+                <option id="second"></option>
+              </select>
+              <p id="note"><i>Note: Once you publish a recipe you cannot edit the public version</i></p>
```
oturner24: Doesn't the public recipe get updated when the user edits the private recipe?

oturner24 (changes requested):

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -16,11 +25,36 @@
...
+    
+    try {
+      Entity publicRecipeEntity = results.asSingleEntity();
+      // publicRecipeEntity will only equal null if there is no entity left after the filters are applied
+      if (publicRecipeEntity != null) {
+        long publicRecipeID = publicRecipeEntity.getKey().getId();
+        Key recipeEntitykey = KeyFactory.createKey("PublicRecipe", publicRecipeID);
+        datastore.delete(recipeEntitykey);
+      }
+    } catch (TooManyResultsException e) {
```
ez-may: So while this is correct error to catch, simply not doing the transaction of deleting the public recipe doesn't seem like the correct result. If for some reason this error were to ever be caught, there would no longer be a private recipe to reference, but there are more than one public recipes with the same name, author, and recipe still in Datastore; this is fatal because the user expects one of them to not exist anymore. I would suggest looking at a different way of approaching the solution, and asking Alex or Olivia for more feedback about whether the areas where this implementation could struggle are worth accounting.

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
```diff
@@ -16,11 +25,36 @@
...
+
+    Query query = new Query("PublicRecipe").setFilter(composFilter);
+    PreparedQuery results = datastore.prepare(query);
+    
+    try {
+      Entity publicRecipeEntity = results.asSingleEntity();
+      // publicRecipeEntity will only equal null if there is no entity left after the filters are applied
+      if (publicRecipeEntity != null) {
+        long publicRecipeID = publicRecipeEntity.getKey().getId();
+        Key recipeEntitykey = KeyFactory.createKey("PublicRecipe", publicRecipeID);
```
ez-may: This step seems like it isn't needed. You do `publicRecipeEntity.getKey()` already. Why not have one line of `datastore.delete(publicRecipeEntity.getKey())` ?

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -29,13 +36,12 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String descriptionResponse = request.getParameter("edit-description").trim();
     String ingredientsResponse = request.getParameter("edit-ingredients").trim();
     String stepsResponse = request.getParameter("edit-steps").trim();
-    // privacy will be used once we give users the option to make their recipes public
-    // String privacy = request.getParameter("privacy");
+    String privacy = request.getParameter("privacy");
```
ez-may, line 32: Nit: follow your variable naming convention. `privacy` -> `privacyResponse`

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -67,17 +78,65 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
         steps = FormHelper.separateByNewlines(stepsResponse);
       }
 
+      // recipe is private and stays private
```
ez-may: What is this comment referencing?

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -67,17 +78,65 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+
+      if (privacy.equals("public")) {
+        recipeEntity.setProperty("published", true);
+      }
+      if (privacy.equals("private")) {
+        recipeEntity.setProperty("published", false);
+      }
+
+      // recipe was private and is made public
+      if (!published && privacy.equals("public")) {
```
ez-may, line 88: What if a recipe is public and was made private?

ez-may (changes requested):

src/main/java/com/google/sps/data/Recipe.java
isamitre: actually I never actually use the Recipe object when doing this so I'm just going to delete the methods

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
isamitre: I'll add a TODO for that
I'm working on the user stuff with Ali right now so I'll get back to it when I'm done
that makes a lot of sense though!

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: will do!

src/main/webapp/pages/UserPage.jsp
isamitre: oh yeah whoops

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
isamitre: yup that makes sense will do

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
isamitre: yeah no I'll just do what you suggested of storing the public recipe ID in the private recipe

src/main/java/com/google/sps/servlets/DeleteRecipeServlet.java
isamitre: actually I'm gong to follow Mathew's suggestion and add a publicRecipeID variable to Recipe

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: I only do the response thing above bc everything else has a chance of being changed if they don't enter any values or leave only spaces. privacy is a dropdown so they can't leave something that has to be replaced by the datastore value

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: line 105

isamitre:
@Wild-Monkey-Coder @oturner24  I made the changes suggested by both of you and now the the private recipe entity stores the public recipe id for easy retrieval/deleting.

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -8,12 +8,14 @@
...
   private ArrayList<String> tags;
+  private boolean published;
   private String authorID;
   private String description;
   private String imageBlobKey;
   private String name;
-  private int popularity;
+  private long popularity;
   private long id;
+  private long publicRecipeID;
```
oturner24, line 18: Now that you're storing the public recipe ID, I don't think you strictly need the published variable anymore. From what I understand from reading your code, if the recipe is not published, publicRecipeID = 0, and if it is published, publicRecipeID != 0. So you could replace your checks of the boolean published with those checks instead.

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -37,22 +40,21 @@ public Recipe(long id, String name, String imageBlobKey, String description, Arr
     this.ingredients = ingredients;
     this.steps = steps;
     this.popularity = popularity;
+    publicRecipeID = 0;
+    published = true;
```
oturner24, line 44: Setting this default to be different than your other constructor seems error prone. There's no actual enforcement that this constructor is going to be used _only_ for the case you intend, so it could be misused in the future. Can you use the constructor below, and pass in those values when you need them?

Also, side note: if the recipe is published, the publicRecipeID should never be 0, right?

src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -8,12 +8,14 @@
...
   private String description;
   private String imageBlobKey;
   private String name;
-  private int popularity;
+  private long popularity;
   private long id;
+  private long publicRecipeID;
 
   // Recipe constructor for the recipe cards for which less information is needed
   public Recipe(long id, String name, String imageBlobKey, String description) {
```
oturner24, line 21: Do you still need this?

oturner24 (changes requested):

src/main/java/com/google/sps/data/Recipe.java
isamitre: that makes sense will do

isamitre: oh actually I do use the published variable to tell if the recipe entity is a PublicRecipe or a PrivateRecipe
bc if it's published and the public == 0 then it's a public recipe else it's a PrivateRecipe

src/main/java/com/google/sps/data/Recipe.java
isamitre: if it is a PublicRecipe, then yes bc it helps me tell between the public/private recipe entities
I could make a recipe type variable? would that be better?

src/main/java/com/google/sps/data/Recipe.java
isamitre: yes that's what Mathew and I are using for all the recipe cards that aren't the ones in user recipes

PR #26, opened 2020-07-22
User session

Aliblas:
I did some clean up in user-auth.js, including adding checks that will prevent ugly (albeit harmless) error logs in the console. Most importantly, user's now have sessions. idToken is stored in HttpSession on sign-in and destroyed on sign-out. This removes the tedium of sending idToken with every request for user actions. it can now be accessed in a servlet via
request.getSession().getAttribute("idToken")


isamitre (approved):

AlexanderEllis (approved):
LGTM

PR #28, opened 2020-07-23
Recipe

isamitre:
converted everything to only have a Recipe entity
the edit and delete buttons work just as well
making recipes public/private is just changing the published variable
this branch would be in place of my EditBothRecipes branch


src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -68,16 +69,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
       recipeEntity.setProperty("steps", steps);
-      recipeEntity.setProperty("imageBlobKey", (String) recipeEntity.getProperty("imageBlobKey"));
+      recipeEntity.setProperty("tags", tags);
+
+      if (privacy.equals("public")) {
+        recipeEntity.setProperty("published", true);
+      }
+      if (privacy.equals("private")) {
+        recipeEntity.setProperty("published", false);
+      }
```
oturner24: If privacy equals some other value besides public and private, can you default it to published = false? It seems like you should either still set the property, or return an error, rather than fail silently.

also nit, but can you make the second if an else if? These 2 statements should be mutually exclusive, so no need to check them both if the first is true.

src/main/webapp/javascript/user-page.js
```diff
@@ -2,29 +2,32 @@
 const NO_PLANNER_RECIPES = "You have not added any recipes to your planner yet.";
 const NO_COOKBOOK_RECIPES = "You have not added any recipes to your cookbook yet.";
 const NO_USER_RECIPES = "You have not uploaded any recipes yet.";
+const USER_RECIPES_DIV = "user-recipes";
 
 // loads all the recipes when the recipe loads
 async function loadRecipes() {
   loadUserRecipes();
+  setUpDivWithNoRecipes("planner-recipes", NO_PLANNER_RECIPES);
+  setUpDivWithNoRecipes("cookbook-recipes", NO_COOKBOOK_RECIPES);
```
oturner24, line 13: maybe save "planner-recipes" and "cookbook-recipes" as constants? To be consistent with USER_RECIPES_DIV, and also so they're easier to reuse if you needed to reuse them anywhere

oturner24 (approved):

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -68,16 +69,22 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
       datastore.put(recipeEntity);
     } catch (EntityNotFoundException e) {
-      // in normal circumstances, this won't happen bc user has not access to id
-      System.out.println("entity not found exception");
+      System.out.println("EditRecipeServlet: Private recipe entity not found with saved recipe id. This should never happen.");
```
ez-may, line 86: Not major, just a thought. I figure that if a user had two tabs open, then deleted one recipe from one tab, went to the second tab, and tried to edit the recipe, you would get this behavior. I think that some simple error handling here, like changing the server response and having the client ask the user to reload the page through an alert could be useful.

ez-may (approved):
LGTM, and looks much easier to maintain!

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: so after the check for public I'll just have an else that sets published false then

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: I tried to automatically reload, but it leads to issues right now since every time I refresh it asks me to log in again
my goal is to reload just the divs but idk if I'll get that to work

PR #29, opened 2020-07-24
Adds author

isamitre:
adds user authentication to UserPage
adds authorID to Recipe entity
lists user recipes by filtering the Recipe entity by the authorID
it does make you re-login every time the page reloads though

### DIFF
--- a/src/main/java/com/google/sps/data/Recipe.java
+++ b/src/main/java/com/google/sps/data/Recipe.java
@@ -13,8 +13,9 @@ public class Recipe {
   private String description;
   private String imageBlobKey;
   private String name;
-  private int popularity;
+//   private int popularity;
   private long id;
+  private long publicRecipeID;
 
   // Recipe constructor for the recipe cards for which less information is needed
   public Recipe(long id, String name, String imageBlobKey, String description) {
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/sps/data/Recipe.java
```diff
@@ -23,16 +23,18 @@ public Recipe(long id, String name, String imageBlobKey, String description) {
...
     tags = new ArrayList<String>();
     published = false;
   }
 
-  public Recipe(long id, String name, String imageBlobKey, String description, ArrayList<String> tags, 
+  public Recipe(long id, String name, String authorID, String imageBlobKey, String description, ArrayList<String> tags, 
       ArrayList<String> ingredients, ArrayList<String> steps, boolean published, long popularity) {
     this.id = id;
     this.name = name;
+    this.authorID = authorID;
```
ez-may, line 37: Nit: are we not using the convention on `authorId` instead of `authorID`?

ez-may (approved):
LGTM

src/main/java/com/google/sps/data/Recipe.java
isamitre: the methods/functions we keep in camelcase (i.e. getAuthorId, getId) but the variable names we said could stay as ID

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -46,14 +48,16 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
+    User user = new User(idToken);
+
     Entity recipeEntity = new Entity("Recipe");
     recipeEntity.setProperty("name", name);
     recipeEntity.setProperty("tags", tags);
     recipeEntity.setProperty("description", description);
     recipeEntity.setProperty("ingredients", ingredients);
     recipeEntity.setProperty("steps", steps);
+    recipeEntity.setProperty("authorID", user.getId());
```
oturner24, line 59: Can you do `recipeEntity.setProperty("authorID", idToken)`? (and not create a User) Or is this id different?

oturner24 (approved):

src/main/java/com/google/sps/servlets/NewRecipeServlet.java
isamitre: the id is different; I have to get it from the idToken
to be honest, I'm just doing what Ali told me; he said to get the id by making a new User

PR #31, opened 2020-07-24
Recipe page now works.

ez-may:
This is the code for redirecting from a recipe card to the recipe page using the recipe ID. There are also functions which handle the creation of HTML elements with IDs which are easy to edit for purposes of CSS.
image

image

### DIFF
--- a/src/main/java/com/google/sps/servlets/IdSearchServlet.java
+++ b/src/main/java/com/google/sps/servlets/IdSearchServlet.java
@@ -1,0 +1,79 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.KeyFactory;
+import com.google.sps.data.Recipe;
+import com.google.sps.util.Utils;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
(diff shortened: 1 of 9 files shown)


isamitre (approved):
LGTM can't wait to see this in action!

src/main/java/com/google/sps/servlets/IdSearchServlet.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** 
+ * Handles requests for a Datastore search with a given ID.
+ **/
+@WebServlet("/retrieve-id")
```
oturner24: nit: "retrieve-id" sounds more like you're getting the ID, rather than retrieving information with a given ID. Maybe change to something like "retrieve-by-id" or "id-search"?

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -50,11 +51,77 @@ createRecipeCard = (divID, recipeInfo) => {
...
+ * search params.
+ */
+async function fillRecipeTemplate() { 
+  const recipeId = (new URL(document.location)).searchParams.get("id");
+
+  // Gets the servlet data for the recipe
+  // TODO: put this code in an if statement to handle recipe not found scenarios
+  let response = await fetch("/retrieve-id?recipeId=" + recipeId.toString());
+  let responseText = await response.text();
+  console.log(responseText);
```
oturner24: Do you want to leave this in, or was it just for debugging? If you're keeping it, consider adding some context, like "Response from Id Search Servlet:" so it's obvious what it is without looking at the code.

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -50,11 +51,77 @@ createRecipeCard = (divID, recipeInfo) => {
...
+  elementsForDiv.forEach(elem => infoDiv.appendChild(elem));
+}
+
+/**
+ * Creates the HTML for some part of the recipe which needs to shown as a list
+ * (ingredients, steps, etc.) and adds it to the given div.
+ * Takes a list and div (not div name) being used.
+ */
+addAsList = (listObj, listDiv, listName) => {
+  // Constrcuts the base of how all item created in the function will be
```
oturner24: nit: typo? constrcuts -> constructs

oturner24 (approved):

src/main/java/com/google/sps/servlets/IdSearchServlet.java
ez-may: Ack and fixed

src/main/webapp/javascript/recipe-rendering.js
ez-may: Ack and fixed, I removed it

src/main/webapp/javascript/recipe-rendering.js
ez-may: Ack and fixed

PR #34, opened 2020-07-27
Search edits

ez-may:
Changes made to be compatible with the Recipe entity refactor.

### DIFF
--- a/src/main/java/com/google/sps/util/TestUtil.java
+++ b/src/main/java/com/google/sps/util/TestUtil.java
@@ -23,7 +23,7 @@ public class TestUtil {
    * Takes in the info for a recipe and returns an entity with set properties based on the recipe object
    */
   public static Entity createRecipeEntity(String entityName, String name, String author, String description, 
-    ArrayList<String> tags, ArrayList<String> ingredients, ArrayList<String> steps, int popularity) {
+    ArrayList<String> tags, ArrayList<String> ingredients, ArrayList<String> steps, boolean published, int popularity) {
     Entity recipeEntity = new Entity(entityName);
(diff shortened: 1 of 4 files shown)


isamitre (approved):
LGTM

src/main/java/com/google/sps/util/TestUtil.java
```diff
@@ -50,8 +51,8 @@ public static Entity createRecipeEntity(String entityName, String name, String a
     for (int i = 0; i < numRecipes; i++) {
       testRecipes[i] = TestUtil.createRecipeEntity(
         entityName, recipeName, author,
-        message + String.format(" I AM RECIPE %d OF MY KIND.", i + offset),
-        tags, ingredients, steps, Integer.valueOf(0));
+        message + String.format(". I AM RECIPE %d OF MY KIND.", i + offset),
```
AlexanderEllis: What is this additional period at the beginning? Slash what is `message` here?

AlexanderEllis (approved):
LGTM (one minor comment)

src/main/java/com/google/sps/util/TestUtil.java
ez-may: The message is meant to be a string saying what test the recipe is from. For example, in the test files I currently have, it says "THIS IS A TEST RECIPE FOR THE SEARCH BY NAME TEST." This goes in the recipe description box and makes it easy to see whether or not the correct recipe data is being used in a search test. The period helps with formatting, since the string from where it's called had no punctuation. But I think I will move it to the actual method, so that formatting in the future makes more sense.

PR #36, opened 2020-07-28
User query object update

ez-may:
Made changes which were proposed in the design doc to update the user query object. UserQuery now has all the methods which are used to construct filters based on parameters, and simplifies the logic of the search servlet. UserQuery should now be robust enough to handle additions of helper methods to organize other filter data and an eventual transition from arraylistst to sets.


Aliblas (approved):
LGTM

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -14,47 +14,128 @@
...
+    // ArrayList which will hold filters for name, authors, tags, and visibility.
+    ArrayList<Query.Filter> searchFilters = new ArrayList<Query.Filter>();
+    
+    // Creates and adds filters to the list if there is data to make the filter
+    if (tagsAndIngredients.size() > 0){
+      searchFilters.add(createTagsFilter());
+    }
+    if (authors.size() > 0) {
+      searchFilters.add(createAuthorsFilter());
+    }
```
oturner24: Use `!authors.isEmpty()`, same for tagsAndIngredients. In general, it's more readable

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -14,47 +14,128 @@
...
+    if (recipeName != null && !recipeName.equals("")) {
```
oturner24, line 82: Use `!recipeName.isEmpty()` (for same reason as above)

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -14,47 +14,128 @@
...
 
-  public ArrayList<String> getAuthors() {
-    return this.authors;
+  /**
+   * Returns a filter for recipes which have all of the tags and ingredients in the collection
+   * present.
+   */
+  private Query.CompositeFilter createTagsFilter() {
+    // Implementation of collection interface methods to handle future change of ArrayList to set
+    Iterator<String> tagsAndIngredientsIterator = tagsAndIngredients.iterator();
```
oturner24: You can use a for each loop, instead of dealing with the iterator. That's what the for each loop does behind the scenes anyway (https://stackoverflow.com/questions/29601970/what-happens-behind-the-scene-when-i-write-for-loop-like-this)

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -14,47 +14,128 @@
...
+    // param to the composite filter constructor
+    ArrayList<Query.Filter> tagsAndIngredientsFilters = new ArrayList<Query.Filter>();
+    
+    // Fills the arraylist with all the filters
+    while (tagsAndIngredientsIterator.hasNext()) {
+      tagsAndIngredientsFilters.add(
+        createSingleTagFilter(tagsAndIngredientsIterator.next())
+      );
+    }
+    return new Query.CompositeFilter(Query.CompositeFilterOperator.AND, tagsAndIngredientsFilters);
```
oturner24, line 131: What if there is only 1 filter in the tagsAndIngredientsFilters list? Do you need to explicitly check for it like you do above?

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -14,47 +14,128 @@
...
   }
 
-  public String getName() {
-    return this.recipeName;
+  /**
+   * Returns a filter for recipes which match one of the authors present in the search
+   */
+  private Query.CompositeFilter createAuthorsFilter() {
+    // Implementation of collection interface methods to handle future change of ArrayList to set
+    Iterator<String> authorIterator = authors.iterator();
```
oturner24: same here about using for each loop

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -14,47 +14,128 @@
...
+    // the composite filter constructor
+    ArrayList<Query.Filter> authorsFilters = new ArrayList<Query.Filter>();
+    
+    // Fills the arraylist with all the filters
+    while(authorIterator.hasNext()) {
+      authorsFilters.add(
+        new Query.FilterPredicate("authorID", Query.FilterOperator.EQUAL, authorIterator.next())
+      );
+    }
+    return new Query.CompositeFilter(Query.CompositeFilterOperator.OR, authorsFilters);
```
oturner24, line 148: Same question here, do you need to check for size == 1?

src/main/java/com/google/sps/servlets/SearchServlet.java
```diff
@@ -44,27 +43,34 @@
...
-    ));
+    // Changes the size of the response
+    final int resultsToReturn;
+    
+    // If the search request came from the main page, we want the behavior to be different
+    if (request.getHeader("Referer").contains("/pages/MainPage.jsp")) {
+      resultsToReturn = 10;
+    } else {
+      resultsToReturn = 10;
+    }
```
oturner24: Can you add a TODO indicating what might eventually be different for the main page?

Also, you can initialize `resultsToReturn = 10`, and then do the if statement and change it if necessary (you'll have to remove the `final` modifier)

oturner24 (changes requested):

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and fixed.

src/main/java/com/google/sps/data/UserQuery.java
ez-may: I actually do not know what happens, nor do I know how to test this behavior. I'm gonna make a check to be sure

ez-may: Fixed it though

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Fixed

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Fixed

src/main/java/com/google/sps/data/UserQuery.java
ez-may: I had isEmpty(), but then when making my code check for the `size() == 1` case, I switched the if statement to be `size() > 1` because I though it was more consistent. Lmk if you don't agree and I'll make the change to `isEmpty()`

src/main/java/com/google/sps/servlets/SearchServlet.java
ez-may: Added the TODO, but I left the resultsToReturn the same. I put it as an uninitialized final value, because I thought it would make it clear that the response value should not change in the code/servlet because we want the user to always get the same number of results.

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -59,11 +59,22 @@ public UserQuery(String query, String tags, String authors) {
     ArrayList<Query.Filter> searchFilters = new ArrayList<Query.Filter>();
     
     // Creates and adds filters to the list if there is data to make the filter
-    if (tagsAndIngredients.size() > 0){
+    if (tagsAndIngredients.size() > 1){
       searchFilters.add(createTagsFilter());
+    } else if (tagsAndIngredients.size() == 1){
```
oturner24: nit: add a space between `)` and `}`

Also, can you add a comment why you're treating size = 1 as a special case? (just so it's recorded and obvious to future code readers)

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -59,11 +59,22 @@ public UserQuery(String query, String tags, String authors) {
...
+      // Use of collection interface methods only to allow for future flexibility
+      String tempTag = tagsAndIngredients.iterator().next();
+      searchFilters.add(createSingleTagFilter(tempTag));
     }
-    if (authors.size() > 0) {
+    if (authors.size() > 1) {
       searchFilters.add(createAuthorsFilter());
+    } else if (authors.size() == 1) {
+      // Use of collection interface methods only to allow for future flexibility
+      String tempAuthor = authors.iterator().next();
```
oturner24: nit: can you change all the variable names with "temp" to more meaningful variable names? Like this could be `firstAuthor` or `singleAuthor`, for example

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -103,37 +114,32 @@ public UserQuery(String query, String tags, String authors) {
...
-    // Implementation of collection interface methods to handle future change of ArrayList to set
-    Iterator<String> tagsAndIngredientsIterator = tagsAndIngredients.iterator();
-    
     // An array list of all the filters for tags/ingredients which can then be passed as a 
     // param to the composite filter constructor
     ArrayList<Query.Filter> tagsAndIngredientsFilters = new ArrayList<Query.Filter>();
     
     // Fills the arraylist with all the filters
-    while (tagsAndIngredientsIterator.hasNext()) {
+    for (String loopTag : tagsAndIngredients) {
```
oturner24: super nit: can you name this just tag? I know it's in a loop but that's not super relevant to what it actually represents.

ditto with loopAuthor below

src/main/java/com/google/sps/servlets/SearchServlet.java
oturner24: I don't fully understand why you're setting it in the if-statements - if you always want it to be the same value anyway, you should define it once as a constant, right?

Aliblas: I personally think we should only ever display twelve recipes or potentially multiples of twelve. It's the only way to have the website always have full rows regardless of whether num-columns is 4, 3, 2, or 1.

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and added

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and changed

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and changed

ez-may:
I added the code for getting telling if the page that made request came from a specific page with a descriptive comment. When I first went looking for a solution to this it wasn't immediately clear to me how to do this, so I wanted to document it somewhere.

oturner24 (approved):

PR #37, opened 2020-07-28
save/unsave from planner/cookbook

isamitre:
recipes can be added to a user's lists from anywhere and removed only from the user page
if it's already been added, then they get an alert informing them so
before removing from a list, they need to confirm that they wish to delete
also changed the icons used for adding/removing from a list


isamitre:
image

src/main/java/com/google/sps/servlets/AddToListServlet.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting recipes. */
+@WebServlet("/add-list")
+public class AddToListServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
```
ez-may: Overall, this get method is lacking in comments, even if it's just an all encompassing one at the beginning of the method. Same goes for some of your servlets below.

src/main/java/com/google/sps/servlets/AddToListServlet.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.sps.data.User;
+import java.io.IOException;
+import java.util.List;
+import java.util.Collections;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for deleting recipes. */
```
ez-may: Old comment? Need to update?

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -20,10 +20,29 @@ createRecipeCard = (divID, recipeInfo) => {
...
+  const name = recipeInfo["name"];
+  const idToken = getIdToken();
+
+  const addToPlannerButton = createElement("button", "", {"class": "card-button bottom more-left fas fa-plus add-to-planner-btn"});
+  addToPlannerButton.appendChild(document.createTextNode(" Planner"));
+  addToPlannerButton.addEventListener('click', () => addToList(id, name, idToken, 'planner'));
+
+  const addToCookbookButton = createElement("button", "", {"class": "card-button bottom more-right fas fa-plus add-to-cookbook-btn"});
+  addToCookbookButton.addEventListener('click', () => addToList(id, name, idToken, 'cookbook'));
+  addToCookbookButton.appendChild(document.createTextNode(" Cookbook"));
```
ez-may: Why are you creating the text node and appending it? `createElement()` already does this if you pass in the text for the object parameter.

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -95,4 +95,50 @@ createImage = (name, blobkey) => {
   imageElement.className = "recipe-card-image";
   imageElement.src = "/serve?blobkey="+blobkey;
   return imageElement;
+}
+
+// lets the user know if they have already added a recipe to a particular list
+// otherwise adds the recipe
+function addToList(id, name, idToken, type) {
+  console.log("addToList", name);
```
ez-may: Do you still need this console.log statement?

ez-may (commented):
Maybe I missed something, but is there a reason that you couldn't combine the AddToListServlet and RemoveFromLisServlet into something like ManageListServlet? Looking at it, there seems to be a lot of code you could combine and then handle with if-statements.
Also, while I definitely don't think this is necessary part of this PR, it does seem a little confusing that a recipe shows up in the cookbook and My Recipes, but only in the cookbook has the correct widget indicating the user to remove that recipe. I don't know if that's an easy fix or not, but I just wanted to point that out.

src/main/webapp/javascript/recipe-rendering.js
isamitre: nope

src/main/webapp/javascript/recipe-rendering.js
isamitre: you're right that was a remnant of when I had a different kind of icon that needed me to do that

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.sps.data.User;
+import java.io.IOException;
+import java.util.List;
+import java.util.Collections;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for adding recipe keys to user's planner/cookbook lists. */
```
ez-may: This servlet also handles deletion, so this comment needs updating.

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+      contains = true;
+    } else {
+      contains = false;
+    }
+
+    response.setContentType("text/html;");
+    response.getWriter().println(String.valueOf(contains));  
+  }
+
+  // adds a recipe key to user's list
```
ez-may: This comment needs updating since it also deletes

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+    String type = request.getParameter("type");
+    String action = request.getParameter("action");
+
+    // creates a key based on the passed in recipe ID
+    Key key = KeyFactory.createKey("Recipe", id);
+    // creates a user based on the passed in user id token
+    User user = new User(idToken);
+
+    // adds created key to appropriate list based on passed in list type
+    if (type.equals("cookbook")) {
```
ez-may, line 64: Could you simplify this by doing something like 
`if ((type.equals("cookbook") || type.equals("planner")) && (action.equals("add") || action.equals("remove")) {do stuff} else { error message}`
This would definitely reduce the number of SOPs you have to do.
Then for the do stuff bracket you could do `if (action.equals("add")){ if (type.equals("planner")) {add planner} else {add cookbook}}`
Then for the remove you do something similar. This last suggestion is definitely out there, but I do think that doing your checks for invalid input at the beginning would help.

ez-may (approved):
LGTM

src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
```diff
@@ -0,0 +1,70 @@
...
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String idToken = request.getParameter("idToken");
+    String type = request.getParameter("type");
+    User user = new User(idToken);
+    String authorID = user.getId();
+
+    List<Key> keys;
+    if (type.equals("cookbook")) {
+      keys = user.getCookbookList();
+    } else if (type.equals("planner")) {
```
oturner24, line 37: Would it make sense to use `equalsIgnoreCase` or convert type to lowercase?

src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
```diff
@@ -0,0 +1,70 @@
...
+      try {
+        Entity entity = datastore.get(key);
+        long id = key.getId();
+        String name = (String) entity.getProperty("name");
+        String description = (String) entity.getProperty("description");
+        String blobkey = (String) entity.getProperty("imageBlobKey");
+
+        Recipe recipe = new Recipe(id, name, blobkey, description);
+        recipes.add(recipe);
+      } catch (EntityNotFoundException e ) {
```
oturner24: nit: remove the space after `e`

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Servlet responsible for adding recipe keys to user's planner/cookbook lists. */
+@WebServlet("/manage-list")
+public class ManageListServlet extends HttpServlet {
+
+  // checks whether the recipe already exists in the user's list
```
oturner24: super nit: for method comments, use `/** ... */` and capitalize the first word (same below)

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+    boolean contains = false;
+
+    // creates a user based on the passed in user id token
+    User user = new User(idToken);
+    // gets the appropriate list based on the passed in list type
+    List<Key> keys;
+    if (type.equals("cookbook")) {
+      keys = user.getCookbookList();
+    } else if (type.equals("planner")) {
+      keys = user.getPlannerList();
```
oturner24, line 35: same question here, re: comparing ignoring case

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+
+/** Servlet responsible for adding recipe keys to user's planner/cookbook lists. */
+@WebServlet("/manage-list")
+public class ManageListServlet extends HttpServlet {
+
+  // checks whether the recipe already exists in the user's list
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // gets values from request
+    long id = Long.parseLong(request.getParameter("id"));
```
oturner24: nit: can you give this a more descriptive name? It's not immediately clear that this is the recipe id (I think) and the idToken is for the user.

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -0,0 +1,89 @@
...
+
+    // creates a key based on the passed in recipe ID
+    Key key = KeyFactory.createKey("Recipe", id);
+    // checks whether the key already exists in list
+    if (keys.contains(key)) {
+      contains = true;
+    } else {
+      contains = false;
+    }
+
```
oturner24: Remove the `contains` variable, and do something like:
`response.getWriter().println(String.valueOf(keys.contains(key));`

src/main/java/com/google/sps/servlets/ManageListServlet.java
oturner24: I agree the if-statements are sort of a lot to look at, but I actually think there's value in having more specific error messages, if you already have to do all these checks anyway.

One idea to simplify the code would be to have a couple helper methods, something like `actionOnCookbook(key, action)` and a similar one for planner. Then this code would look like:
```
if (type.equals("planner")) {
   actionOnPlanner(key, action);
} else if (type.equals("cookbook")) {
   actionOnCookbook(key, action);
} else {
   // print "wrong type" error
}
```
and then the inner if statements would be in the helpers. You could also split it up the opposite way, and have `addToListType(key, type)` and `removeFromListType(key, type)`. These aren't huge wins but it would reduce the nesting which can help with readability.

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -95,4 +93,70 @@ createImage = (name, blobkey) => {
...
+      if (action == "add") {
+        // if the user is trying to add a recipe not already in the list
+        const params = new URLSearchParams();
+        params.append("action", action);
+        params.append("id", id);
+        params.append("idToken", getIdToken());
+        params.append("type", type);
+        fetch("/manage-list", {method: "POST", body: params});
+        if (document.URL.includes("UserPage")) {
+            location.reload();
```
oturner24: nit: this indentation looks too far right

oturner24 (approved):

src/main/java/com/google/sps/servlets/ManageListServlet.java
isamitre: will do!

src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
isamitre: done

PR #38, opened 2020-07-29
Pretty main page

Aliblas:
This branch is built form Mathew's userQueryUpdate branch, so it would be best to merge that first.
I wanted to make recipes resize a to fill their parent div, my efforts failed so more on that later.
Important note: SearchServlet and IdSearchServlet work for testData Recipes, but they can't seem to find Recipes uploaded from the user page.
image


src/main/java/com/google/sps/servlets/NewRecipeServlet.java
```diff
@@ -59,6 +59,7 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     recipeEntity.setProperty("authorID", user.getId());
     recipeEntity.setProperty("popularity", 0);
     recipeEntity.setProperty("published", false);
+    recipeEntity.setProperty("popularity", 0);
```
ez-may: Remove this , already on line 60.

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.IOException;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/settings")
+public class UserSettingsServlet extends HttpServlet {
+  // The only reason this class exists rn is to save this regular expression I made.
+  // It checks that user names are alphanumeric and (5 - 25) characters.
```
ez-may: I understand this, but I think you should put this in another class's comments or as a function in a util? Having an entire servlet to hold a comment doesn't seem like the right approach.

ez-may (changes requested):
Change entire servlet for one comment

ez-may (approved):
LGTM

src/main/webapp/components/topbar.jsp
```diff
@@ -1,20 +1,24 @@
 <!--To use this top bar in other parts of the application, use the jsp @include directive
   in another jsp file to merge this html into that file.-->
 <div id="top-bar">
+  <div id-"logo">
```
AlexanderEllis: Should be `id=`

src/main/webapp/javascript/search.js
```diff
@@ -34,7 +34,15 @@ function redirectSearchPage() {
 /**
```
AlexanderEllis, line 34: Are this and MainPage.jsp from userQueryUpdate?

AlexanderEllis (approved):
CSS LGTM (happy to take another look after we merge userQueryUpdate branch)

src/main/webapp/javascript/search.js
ez-may: I can step in here, but I'm not entirely sure I remember. UserQueyUpdate was more focused on the updates to the servlet and such, but there was a much smaller PR earlier today (Friday 31, July) that was search redirecting

PR #39, opened 2020-07-30
Popularity

isamitre:
popularity of a recipe increases by one when it is added to the cookbook


src/main/java/com/google/sps/servlets/EditRecipeServlet.java
```diff
@@ -75,15 +76,17 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
       recipeEntity.setProperty("ingredients", ingredients);
       recipeEntity.setProperty("steps", steps);
       recipeEntity.setProperty("tags", tags);
+      recipeEntity.setProperty("popularity", popularity);
       recipeEntity.setProperty("published", false);
 
       if (privacy.equals("public")) {
         recipeEntity.setProperty("published", true);
+        recipeEntity.setProperty("popularity", popularity++);
```
ez-may: What would happen if a user continually makes their recipe public, then private, and so on? Would they be able to make their recipe infinitely popular?

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -55,31 +59,68 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     } else if (action.equals("remove")) {
       user.removeCookbookKey(key);
+      decreasePopularity(key);
     } else {
       System.out.println("ManageListServlet: invalid action " + action);
     }
   }
-  public void actionOnCookbook(User user, Key key, String action) {
+
+  public void increasePopularity(Key key) {
```
ez-may: My only  comment here is I would've used a general method like `updatePopularity` which could take a positive or negative int and would have a lot of the shared logic of the increase/decrease popularity functions, and then have your functions make call to that function. Just an as abstraction since the only apparent difference is the value change and the logic of `if (popularity < 0) { popularity = 0}`

src/main/java/com/google/sps/servlets/EditRecipeServlet.java
isamitre: wait no idk why I did that that doesn't make sense
I'm going to remove that

src/main/java/com/google/sps/servlets/ManageListServlet.java
isamitre: done

ez-may (approved):
LGTM

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -55,31 +59,58 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
     } else {
       System.out.println("AddToListServlet: invalid type " + type);
     }
   }
 
-  public void actionOnPlanner(User user, Key key, String action) {
+  public void actionOnCookbook(User user, Key key, String action) {
     if (action.equals("add")) {
       user.addCookbookKey(key);
+      updatePopularity(key, "increase");
```
AlexanderEllis: You should make "increase" and "decrease" be constants (even though they're referenced only twice, it'll keep it cleaner).

That being said if you wanted to skip the actions and the string checks, you could have the second param be a bool named `increase` where true -> popularity++ and false -> popularity-- (or something similar). Up to you though!

src/main/java/com/google/sps/servlets/ManageListServlet.java
```diff
@@ -55,31 +59,58 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
     String type = request.getParameter("type").toLowerCase();
     String action = request.getParameter("action").toLowerCase();
 
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
AlexanderEllis: Are you using this `datastore` variable anywhere in this `doPost`? I think you can remove it.

AlexanderEllis (approved):
LGTM w/ few comments 👍

src/main/java/com/google/sps/servlets/ManageListServlet.java
isamitre: cool followed the second suggestion!

PR #42, opened 2020-07-31
Search redirecting

ez-may:
Very small changes made to the search.js file to allow for redirecting when a query is made in the search bar.

### DIFF
--- a/src/main/webapp/javascript/search.js
+++ b/src/main/webapp/javascript/search.js
@@ -4,8 +4,6 @@
  * TODO: implement redirect to search page
  */
 async function search() {
-    let userQuery =  document.getElementById("search-box").value;
-    document.getElementById("content").innerHTML = "";
     // Uncomment to initialize test data for search testing
     /*let tempForm = new FormData();
     tempForm.append("action", "initialize");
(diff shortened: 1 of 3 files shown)


isamitre (approved):
LGTM

src/main/webapp/javascript/search.js
```diff
@@ -16,13 +14,23 @@ async function search() {
...
     let recipeList = JSON.parse(responseText);
 
     recipeList.forEach(elem => createRecipeCard("content", elem));
 }
 
+/**
+ * Redirects to the search page with the query in the URL search params.
+ */
+function redirectSearchPage() {
+  let userQuery =  document.getElementById("search-box").value;
```
AlexanderEllis: I think this can be `const` instead of `let` since you aren't altering it at all

src/main/webapp/javascript/search.js
```diff
@@ -16,13 +14,23 @@ async function search() {
     };
     fetch("/adminAuth_test_congifureTestData?", myInit);*/
 
-    let response = await fetch("/search?query=" + userQuery);
+    const searchQuery = (new URL(document.location)).searchParams.get("query");
+
+    let response = await fetch("/search?query=" + searchQuery);
```
AlexanderEllis: `let` -> `const` for these three since you aren't making changes to them?

AlexanderEllis (approved):
LGTM!

src/main/webapp/javascript/search.js
ez-may: Ack and changed

PR #46, opened 2020-08-03
Not logged in

isamitre:
This PR will
- exclude private recipes from being seen in planner/cookbook lists
- delete keys from cookbook/planner lists if an entity with that key is not found (meaning the recipe was deleted by its creator)
- automatically (without refreshing) remove a recipe card from the planner/cookbook sections if the recipe is deleted 
- have a default of a plus sign on the add to list buttons on a recipe card
- have a separate function that changes a recipe card's add to list button to a checkmark if it has already been added to that list
- prompt the user to log in if they click on an add to list button (only seems to work on mainpage rn)

(NOTE: that separate function in the penultimate bullet point assumes the user is logged in so rn it can only be used in the user page)


src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
```diff
@@ -44,19 +45,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
     
     List<Recipe> recipes = new ArrayList<>();
+    List<Key> deletedKeys = new ArrayList<>();
```
ez-may, line 50: They way I understood this is that deleted keys are the datastore entities a user had which no longer exist, but this also extends to public recipes. I would maybe add a comment to explain this because I didn't understand until I read more of your code further down.

src/main/webapp/javascript/user-page.js
```diff
@@ -60,14 +58,17 @@ async function loadTypeRecipes(type) {
 
   let recipesDivID = type + "-recipes";
   document.getElementById(recipesDivID).innerHTML = "";
+
+  let other;
```
ez-may: Nit: needs comment here. I understand how you're using it, but it's no immediately clear

ez-may (approved):
LGTM I think

src/main/webapp/javascript/user-page.js
isamitre: will do

src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
```diff
@@ -44,19 +45,35 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    }
+
+    for (Key key: deletedKeys) {
+      // TODO: inform user how many recipes were deleted
+      if (type.equals("cookbook")) {
+        user.removeCookbookKey(key);
+      } else if (type.equals("planner")) {
+        user.removePlannerKey(key);
+      } else {
+         System.out.println("ListTypeRecipesServlet: invalid type");
```
oturner24: Instead of iterating through the deleted keys and maintaining a separate array list, can you move this logic into the catch block?

Also, can you add the type to the error message that's printed? Like you do in ManageListServlet.

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -14,20 +14,19 @@ createRecipeCard = (divID, recipeInfo) => {
...
+  addToPlannerButton.addEventListener('click', () => manageList("add", id, name, "planner"));
 
-  const addToCookbookButton = createElement("button", " Cookbook", {"class": "card-button action-button bottom more-right fas fa-plus add-to-cookbook-btn"});
-  addToCookbookButton.addEventListener('click', () => manageList("add", id, name, idToken, 'cookbook'));
+  const addToCookbookButton = createElement("button", " Cookbook", {"class": "card-button action-button bottom more-right add-to-cookbook-btn fas fa-plus"});
+  addToCookbookButton.addEventListener('click', () => manageList("add", id, name, "cookbook"));
 
   const removeFromPlannerButton = createElement("button", " Planner ", {"class": "card-button action-button bottom more-left fa fa-remove remove-from-planner-btn"});
-  removeFromPlannerButton.addEventListener('click', () => manageList("remove", id, name, idToken, 'planner'));
+  removeFromPlannerButton.addEventListener('click', () => manageList("remove", id, name, 'planner'));
```
oturner24: nit: change 'planner' to use "" to be consistent with above. Same for 'cookbook' below

src/main/webapp/javascript/user-page.js
```diff
@@ -61,13 +59,17 @@ async function loadTypeRecipes(type) {
   let recipesDivID = type + "-recipes";
   document.getElementById(recipesDivID).innerHTML = "";
   
+  // other will be used to set up the icon of the other add to list button
+  let other;
```
oturner24: Seems like you could get rid of this variable and following if statement, and move the logic below where you're calling changeIcons, like:
```
if (type == "planner") {
   changeIcons(recipesDiv, "cookbook")
} else {
   changeIcons(recipesDiv, "planner")
}
```
This way it's more flexible to future changes and you avoid storing an extra variable

src/main/webapp/javascript/user-page.js
```diff
@@ -85,18 +87,39 @@ async function loadTypeRecipes(type) {
...
         removeFromListButton.style.display = "inline-block";
       }
+      // sets up the icon of the other add to list button to a checkmark if it is contained
+      changeIcons(recipesDiv, other);
     }
   }
 }
 
+function changeIcons(recipesDiv, type) {
+    console.log("changeIcons");
```
oturner24: remove this? Or if you want to keep it for debugging purposes, looks like the indentation is too far right

src/main/webapp/javascript/user-page.js
```diff
@@ -85,18 +87,39 @@ async function loadTypeRecipes(type) {
...
+  // TODO: check if they're logged in
+  const recipeIDs = recipesDiv.getElementsByClassName("recipe-id");
+  const addToListButtons = recipesDiv.getElementsByClassName("add-to-" + type + "-btn");
+  const removeFromListButtons = recipesDiv.getElementsByClassName("remove-from-" + type + "-btn");
```
oturner24: It doesn't look like you're using `removeFromListButtons`?

oturner24 (approved):

src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
isamitre: like delete the key in the catch block? I tried to do that and it threw an error

isamitre: will do the second!

src/main/webapp/javascript/user-page.js
isamitre: lol that makes a lot of sense

src/main/webapp/javascript/user-page.js
isamitre: you're right that was for debugging

src/main/webapp/javascript/user-page.js
isamitre: remnant from the copy paste
I'll delete it

PR #47, opened 2020-08-04
Icons Change Outside Of UserPage

isamitre:
shoutout to Ali for making the authentication stuff work!
when not logged in, the add to list buttons have a plus icon
when logged in, it checks if the recipe is contained in the user's list and changes the plus icon to a checkmark if it is 
whenever a recipe is added to a list, the page reloads so that the icon gets changed

I Ali, have included a CL to explain some changes I made.
All user auth script src and css tags were moved to global dependencies.
User listeners are no longer assigned in the Google platform API script's onload parameter.
Global user listener functions userChanged and signInChanged are assigned in recipe-rendering.js and user-page.js for their respective pages.
onload functions that modify the DOM have been moved to the end of body (assures completion before listener events).


src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -1,3 +1,19 @@
+// Set listeners for rendering based on user status,
+// except for User Page, which has special listeners.
+if (document.title != "My Recipes") {
+  // Listener that triggers when a new or different user signs in.
+  let userChanged = setIcons;
+
+  // Listener that triggers when sign-in status changes (but not when user changes).
+  let signInChanged = function(signedIn) {
```
ez-may: What is this meant to do besides console log? Do you need to include this code in the master branch?

src/main/webapp/javascript/recipe-rendering.js
```diff
@@ -1,3 +1,19 @@
+// Set listeners for rendering based on user status,
+// except for User Page, which has special listeners.
+if (document.title != "My Recipes") {
+  // Listener that triggers when a new or different user signs in.
+  let userChanged = setIcons;
```
ez-may: Is this supposed to be `setIcons -> setIcons()`?

src/main/webapp/javascript/user-page.js
```diff
@@ -2,9 +2,16 @@
...
 const USER_RECIPES_ID = "user-recipes";
 
-// Can be passed to API onload to guarantee auth2 is loaded before execution.
-function initUserPageGoogleSignin() {
-  initGoogleUserWithListener(loadRecipes);
+// Listener that triggers when a new or different user signs in.
+var userChanged = loadRecipes;
+
+// Listener that triggers when sign-in status changes (but not when user changes).
+var signInChanged = function(signedIn) {
```
ez-may, line 9: Same as above, do you need to keep this?

ez-may (approved):
LGTM I think

src/main/webapp/javascript/recipe-rendering.js
Aliblas: I inserted those listeners as placeholders for code I thought might be added by someone else. In this case, I expected that we would want to detect signout and reset all the icons to default.

isamitre: oh okay got it yeah I can do that

src/main/webapp/javascript/recipe-rendering.js
Aliblas: userChanged is the global listener function user in the user-auth file. Here it is set to reference setIcons.

ez-may (approved):

src/main/java/com/google/sps/servlets/ListTypeRecipesServlet.java
```diff
@@ -29,7 +30,14 @@
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String idToken = request.getParameter("idToken");
     String type = request.getParameter("type").toLowerCase();
-    User user = new User(idToken);
+    User user;
+    // if user is not logged in, redirects back to Home
+    try {
+      user = new User(idToken);
+    } catch (IllegalArgumentException e) {
+       response.sendRedirect("/pages/MainPage.jsp");
```
oturner24: nit: indent looks off here

oturner24 (approved):

PR #48, opened 2020-08-04
Tag and Author searching

ez-may:
The Search feature now handles searching by tags and authors. The search page has text-areas for adding authors or tags and updating the search query.
image


src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -79,8 +71,19 @@ public UserQuery(String query, String tags, String authors) {
...
+      searchFilters.add(createTagsFilter());
+    } else if (tagsAndIngredients.size() == 1) {
+      // We check for size equal to 1 because we cannot create a composite filter when we only 
+      // have one filter to pass. The same implementatoin is used for the author filters
+      // Use of collection interface methods only to allow for future flexibility
+      String tempTag = tagsAndIngredients.iterator().next();
+      searchFilters.add(createSingleTagFilter(tempTag));
+    }
+
+    if (!recipeName.equals("null") && !recipeName.equals("")) {
```
ez-may: Is this strange? Yes. Does it work? Yes. Did I try `recipeName != null`? Yes, and it did not work. I had another friend in STEP help me try to figure it out and our closest guess is that when the client sends a fetch request with a param with no values, it decides to send a string version of the value "null" that the server can use. That's why we get this behavior (where the first null is line 88): 
Screenshot 2020-08-03 at 7 36 22 PM

isamitre (approved):
LGTM

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -147,4 +152,17 @@ public UserQuery(String query, String tags, String authors) {
     }
     return new Query.CompositeFilter(Query.CompositeFilterOperator.OR, authorsFilters);
   }
+
+  /**
+   * Takes in a string which ha commas and returns a collection (currently an arrayList<string>)
```
oturner24: nit: typo "ha" -> "has"

src/main/java/com/google/sps/data/UserQuery.java
```diff
@@ -147,4 +152,17 @@ public UserQuery(String query, String tags, String authors) {
...
+   * which contans all the substrings which were separated by commas. Can take a null parameter
+   * and empty string, and should return a collection with no elements in both cases. 
+   */
+  private ArrayList<String> convertToCollection(String params) {
```
oturner24: nit: maybe a better name would be `convertToArrayList` to be precise?

src/main/webapp/javascript/search.js
```diff
@@ -20,17 +20,69 @@ async function search() {
...
+redirectSearchPage = () => {
   const userQuery =  document.getElementById("search-box").value;
   window.location.assign("/pages/Search.jsp?query=" + userQuery);
 }
 
+/**
+ * This function is the onclick function for the Update button in the tag box.
+ * This function does the following (with helper functions):
+ * Gets the text from text areas -> seperates everything by commas and formats
+ * the sring list -> makes new fetch request for query -> updates the query
```
oturner24: nit: typo "sring" -> "string"

oturner24 (approved):

src/main/java/com/google/sps/data/UserQuery.java
ez-may: I was hesitant since the idea was eventually this would be converted to a set, so I didn't want to have to change the name everywhere, but since the internship is almost over and I don't have time, I'll go ahead and change it.

src/main/java/com/google/sps/data/UserQuery.java
ez-may: Ack and fixed

PR #49, opened 2020-08-04
Search and Tag-Box Styling

ez-may:
The CSS for the search and tag box has been completed building off the CSS from other files. The tag box was implemented to be sticky so users could scroll through options and then edit the search further down.
image
image


Aliblas (approved):
LGTM, I was wondering how the vw based resizing works with the Search results not taking up 80% of the page the way they do in the other pages. I recall not everything resized to fit properly when I changed it previously.
EDIT: this is Isabel. I'm going to make a PR to change the widths of recipe cards to percentages and keep the heights of images as vw so maybe it should work?

oturner24 (approved):

PR #50, opened 2020-08-05
Name change settings

Aliblas:
I just made a whole big interface for editing your display name.
Not sure if we'll get around to actually displaying author names by their recipes, but I also wrote a helper method for fetching them with data store for a given authorID. It is in utils/Users.java.
It might be more efficient to get author names with some datastore indexing magic, rather than fetching the whole User entity, but I don't know indexing magic.
edit: author name appears in Recipe objects and will fetch to front end.

CL:
Default display name is anonymous to respect privacy.
settings page shows current display name, allows submission of new name.
settings servlet checks and allows specific characters and patterns.
Front end gives feedback based on name submission.
settings link added to user dropdown.
image

### DIFF
--- a/src/main/java/com/google/sps/servlets/UserSettingsServlet.java
+++ b/src/main/java/com/google/sps/servlets/UserSettingsServlet.java
@@ -1,0 +1,59 @@
+package com.google.sps.servlets;
+
+import com.google.appengine.api.blobstore.BlobInfo;
+import com.google.appengine.api.blobstore.BlobInfoFactory;
+import com.google.appengine.api.blobstore.BlobKey;
+import com.google.appengine.api.blobstore.BlobstoreService;
+import com.google.appengine.api.blobstore.BlobstoreServiceFactory;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.images.ImagesService;
+import com.google.appengine.api.images.ImagesServiceFactory;
+import com.google.sps.data.User;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/settings")
+public class UserSettingsServlet extends HttpServlet {
+  // This regex checks that user names are alphanumeric and (5 - 25) characters.
+  // ([^\s\d]( ?){1}){5,25}
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    System.out.println("Settings fetch\n\n");
+    response.setContentType("application/json");
+    String idToken = (String) request.getParameter("idToken");
+    User user = new User(idToken);
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    System.out.println("Settings post\n\n");
+    String idToken = (String) request.getParameter("idToken");
+    if (idToken == null ) {
+        System.out.println("idtoken was null\n");
+        return;
+    }
+
+    User user = new User(idToken);
+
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/sps/servlets/UserSettingsServlet.java
```diff
@@ -0,0 +1,82 @@
...
+        json.addProperty("message", "Invalid name input");
+      }
+      json.addProperty("success", false);
+    }
+
+    response.setContentType("application/json");
+    response.getWriter().print(json.toString());
+  }
+
+  boolean hasRepeatingSpaces(String word) {
```
isamitre: can't you just do indexOf("  ")? if it returns -1 then there aren't any repeating spaces

isamitre (dismissed):
LGTM except for my comment about indexOf
a picture would be nice too

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
Aliblas: I don't think that works just from chedcking indexOf documentation. The function seams to work so I think I'll stick with it.

isamitre (approved):

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String idToken = (String) request.getParameter("idToken");
+    User user = new User(idToken);
+    JsonObject json = new JsonObject();
+
+    String nameInput = (String) request.getParameter("nameInput");
+    nameInput.trim();
```
oturner24: trim() returns a modified String, it doesn't modify it in place. I think you could combine this all in one if you wanted:
`String nameInput = ((String) request.getParameter("nameInput")).trim();`

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+// So far this servlet allows fetching and uploading a display name.
+@WebServlet("/settings")
+public class UserSettingsServlet extends HttpServlet {
+  // This regex checks that user names are alphanumeric and (5 - 25) characters.
+  static final String namePattern = "^[\\w_ -]{5,25}";
```
oturner24: nit: can you use all caps for constant names?
https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
```diff
@@ -0,0 +1,82 @@
...
+    } else {
+      if (!nameInput.matches(namePattern)) {
+        json.addProperty("message", "Invalid character in requested name");
+      } else if (hasRepeatingSpaces(nameInput)) {
+        json.addProperty("message", "Consecutive spaces not allowed.");
+      } else {
+        json.addProperty("message", "Invalid name input");
+      }
+      json.addProperty("success", false);
+    }
```
oturner24: Up to you, but you could make this more efficient by doing the checks just once:
```
if (!nameInput.matches(namePattern) {
   // set the properties you need
} else if (hasRepeatingSpaces(nameInput) {
   // set the properties you need
} else {
   // here you know it's valid, so you can go ahead and set the name
}

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
oturner24: Is this checking if there are 2 or more spaces in a row in the string? Could you use `word.contains("  ")` instead? (where there are 2 spaces between the quotation marks)

oturner24 (approved):

src/main/java/com/google/sps/servlets/UserSettingsServlet.java
Aliblas: Oh that works, appalling that I didn't think of that.

PR #51, opened 2020-08-05
Seamless

isamitre:
image
image


ez-may (approved):
LGTM, but I think the tag box background should be something with a little more contrast.

Aliblas (dismissed):
LGTM and nicely alphabetized.

oturner24 (approved):
LGTM thanks for including screenshots

PR #52, opened 2020-08-06
Shopping list

Aliblas:
I created a servlet for fetching a non-repeating set of ingredients based on planner recipes. Displays in user page using modal by Isabel and list-printer by Mathew, shoutout -> them.
CL:
Modified addAsList to optionally print ordered or unordered, the new param is optional so this won't break old code.
Created servlet for concatenating ingredients in planner recipes, and returning json array with no repeats.
Added modal on UserPage for displaying shopping list.
Some new code for rendering in user-page.js
I also changed the "Add Recipe" button to "Create Recipe" which I think fits better.
The button:
image
The modal:
image
image


isamitre (approved):
LGTM

oturner24 (approved):
LGTM!

